PR #1, opened 2025-07-24
Initial standalone TTSIM

hmgaudecker:

Extract the ttsim portion from the GETTSIM history.


src/ttsim/__init__.py
MImmesberger, line 1: Doesn't matter much because we will expose them via GETTSIM, but should we add the TTSIM
Object types here? Also `plot_tt_dag`?

`plot_interface_dag` for sure I think

src/ttsim/interface_dag.py
```diff
@@ -0,0 +1,591 @@
...
+    expected_structure = MainTarget.to_dict()
+    # Remove existing top-level elements that are None, these will be calculated.
+    dict_inputs = {
+        k: v.to_dict() if isinstance(v, MainArg) else v
+        for k, v in inputs.items()
+        if v is not None and k in expected_structure
+    }
+    qname_inputs = {}
+    # Special treatment for orig_policy_objects.root because we do not list it in
+    # `MainTarget` so as not to confuse users of GETTSIM, where it is set.
```
MImmesberger, line 164:
```suggestion
    # `MainTarget` so as not to confuse users of TTSIM, where it is set.
```

MImmesberger (approved):
Cool, looks good to me! Should we remove the `documentation` template from the Github Issue templates? There won't be a documentation anytime soon.

src/ttsim/__init__.py
hmgaudecker: Will do that in a different PR.

PR #3, opened 2025-07-24
Renamings

hmgaudecker:
To make it clearer and easier for users, we do the following:
- `tt_dag_elements` -> `tt`. This is used all the time when modifying an environment. At the same time, the namespace is large, so we don't want to clutter the ttsim/gettsim namespace
- Just one typing module. `tt.typing` only had one element left.
- `interface_dag` -> `main`. Just had historical reasons.
- Same for some other stuff that is not user-relevant (content of `interface_dag_elements.__init__` now is in `main_target` etc.)

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -34,7 +34,7 @@ jobs:
             environment: py311, py312
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


MImmesberger (approved):
You fixed all my comments in your latest commits, so looks good to me now :smile:

PR #6, opened 2025-07-24
Fail if the leaf name of an object in the policy environment differs from the last element of the path

MImmesberger:

This adds a modifies the `policy_environment_invalid` function such that it raises an error if the object's leaf name differs from the name of the last path element.

@hmgaudecker What I didn't think about when I agreed to add this, is that parameter would need a leaf name too. Having the experience from writing the notebook I think this would be pretty annoying, especially for small changes, e.g. changing a scalar parameter. Maybe we should just throw a warning?


hmgaudecker (approved):
Very nice, thanks a lot! 

Please have another look as I made quite some changes!

MImmesberger:
Much better, thanks!

PR #9, opened 2025-07-25
Raise error when passing data but no targets

### LINKED ISSUE #4: BUG: Cannot create template when giving data, but no targets
### Bug description

Because we just take all `ColumnFunction`s as targets when the user does not specify targets via `main`, but we remove `ColumnFunction`s from the specialized environment when certain functions are overwritten via data, we get a `dags.exceptions.MissingFunctionsError` when creating a template.

Same should be true when computing TT instead of a template (but the template bug is probably more common).

Test to reproduce:

```python
def test_mettsim_template_with_inputs(backend, xnp):
    main(
        main_target="templates__input_data_dtypes",
        policy_date_str="2025-01-01",
        input_data={
            "tree": {
                "p_id": xnp.array([4, 5, 6]),
                "payroll_tax": {"amount_y": xnp.array([1, 2, 3])},
            }
        },
        orig_policy_objects={"root": Path(__file__).parent.parent / "mettsim"},
        backend=backend,
    )
```

hmgaudecker:

- Fix #4 
- Had to change quite a few tests which should have failed for other reasons, but where this error came first now.
- Update pixi on GHA

### DIFF
--- a/src/ttsim/main.py
+++ b/src/ttsim/main.py
@@ -187,7 +187,7 @@ def _harmonize_inputs(inputs: dict[str, Any]) -> dict[str, Any]:
 
 _MSG_FOR_MISSING_TT_TARGETS = """When providing data, `tt_targets` must be provided.
 
-The reason is ambiguity what you are after when overriding some column
+This is because it's ambiguous what you intend to compute when overriding some column
 that could be computed from primitives.
 
 A simple way out is to first obtain all possible `tt_targets`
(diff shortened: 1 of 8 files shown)


src/ttsim/main.py
```diff
@@ -181,6 +185,26 @@ def _harmonize_inputs(inputs: dict[str, Any]) -> dict[str, Any]:
     return {k: v for k, v in qname_inputs.items() if v is not None}
 
 
+_MSG_FOR_MISSING_TT_TARGETS = """When providing data, `tt_targets` must be provided.
+
+The reason is ambiguity what you are after when overriding some column
```
MImmesberger, line 190:
```suggestion
This is because it's ambiguous what you intend to compute when overriding some column
```

MImmesberger (approved):

PR #11, opened 2025-07-25
Fix several bugs in `plot_tt_dag`

MImmesberger:

@JuergenWiemers pointed out that some nodes are missing when calling `plot_tt_dag`.

It turned out that there were two bugs:
1. When using `NodeSelector`, we didn't add the "main" node to the DAG (`nx` doesn't do so automatically)
2. We still had an `include_other_objects` with default `False` in there from an earlier implementation of `evaluation_x`, `policy_x`, `xnp`, `dnp` and friends. This caused `policy_inputs` and some other nodes to be thrown out of the DAG. I removed the argument completely, we don't need it anymore.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -40,7 +40,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: prefix-dev/setup-pixi@v0.8.13
         with:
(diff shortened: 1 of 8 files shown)


hmgaudecker (approved):
Thanks for the speedy fix!

PR #12, opened 2025-07-25
Turn on `fail_if` nodes when running policy tests

MImmesberger:

I tried both versions (with/without fail nodes) using GETTSIM's tests with the py13 environment and backend numpy. 52 vs 53 seconds, so should be no big deal turning it on.


hmgaudecker (approved):
Thanks for speed-testing!

PR #13, opened 2025-07-25
Put plotting functions in `ttsim.plot.dag` namespace

MImmesberger:

As suggested [here](https://github.com/ttsim-dev/gettsim/pull/1056#pullrequestreview-3056621461).


hmgaudecker (approved):
Wonderful! I just renamed a function in `shared`, still had the old namespacing (sort-of).

PR #17, opened 2025-07-26
Add type for sparse dict with integer keys param

MImmesberger:

Adds the param type `sparse_to_consecutive_int_lookup_table` alongside of `convert_sparse_to_consecutive_int_lookup_table`.


hmgaudecker (approved):
Very nice, thanks! Made some cosmetic changes right in the source and fixed the failing tests -- seems like ruff got a chance to mess with the vectorisation tests during your renamings :laughing:

MImmesberger:
Thanks for finishing it! I was so confused about the sudden vectorisation errors that I took a 3 hour walk instead of trying to fix it :smiley:

> I had second thoughts about the naming

Yes I didn't like my version either, yours is much better. I don't have a better suggestion right now, I don't even think it's that bad.

MImmesberger:
Maybe: `sparse_to_consecutive_int_lookup_table`? Makes it a bit clearer that this parameter is a sparse version of the `consecutive_int_lookup_table` parameter.

PR #23, opened 2025-07-28
Remove orphaned nodes from tt dag

MImmesberger:

The TT DAG has orphan nodes when not selecting a target via `NodeSelector`. 

- Those nodes are:  `evaluation_x`, `policy_x` for `x \in ["year", "month", "day", "date"]`
- They are part of the DAG because they are `policy_inputs`
- They are orphaned if they are not relevant for other `ColumnObject`s. This happens quite often because they usually inform `param_function`s (in GETTSIM), not `policy_function`s.

If this is the case, this PR removes the explicitly from the plotting DAG.


mettsim/middle_earth/housing_benefits/amount.py
```diff
@@ -1,15 +1,32 @@
...
     else:
         return 0
+
+
+@param_function()
+def max_amount_m_fam(
+    policy_year: int,
+    max_amount_m_fam_by_policy_year: ConsecutiveIntLookupTableParamValue,
+) -> float:
+    return max_amount_m_fam_by_policy_year.look_up(policy_year)
```
hmgaudecker, line 32: I'd think it is very unlikely that we'll ever want to have a function like this. For this sort of thing, we have the

```
  1900-01-01:
    value: 2000.0
  1901-01-01:
    value: 2001.0
```
mechanism, right?

If you only want it as a test, better use `evaluation_year` -- otherwise people may take it as a blueprint...

src/ttsim/plot/dag/tt.py
```diff
@@ -161,8 +154,41 @@ def _get_tt_dag_with_node_metadata(
...
+        "policy_day",
+        "policy_date",
+        "evaluation_year",
+        "evaluation_month",
+        "evaluation_day",
+        "evaluation_date",
+    ]:
+        if (
+            x in selected_dag.nodes()
+            and not list(selected_dag.predecessors(x))
```
hmgaudecker, line 185:
```suggestion
            and not set(selected_dag.predecessors(x))
```
Sorry, purist here..

src/ttsim/plot/dag/tt.py
```diff
@@ -161,8 +154,41 @@ def _get_tt_dag_with_node_metadata(
...
+            and not list(selected_dag.successors(x))
```
hmgaudecker, line 186:
```suggestion
            and not set(selected_dag.successors(x))
```

tests/test_plot_dag.py
```diff
@@ -310,3 +309,18 @@ def test_params_are_removed_from_dag(include_params, expected_nodes):
         include_params=include_params,
     )
     assert set(dag.nodes()) == set(expected_nodes)
+
+
+def test_orphaned_policy_inputs_are_removed_from_dag():
```
hmgaudecker, line 314:
```suggestion
def test_orphaned_dates_are_removed_from_dag():
```
We don't want to remove all orphans, right?

hmgaudecker (approved):
Very nice, thanks!

mettsim/middle_earth/housing_benefits/amount.py
MImmesberger: To be honest, I agree. But this contradicts what we did/argued in [this PR](https://github.com/ttsim-dev/gettsim/pull/1044#pullrequestreview-3045607808). We cannot use `evaluation_x` anymore for `param_function`s, only `policy_x` works. And this was intentional.

MImmesberger: Context is this GETTSIM function:

```python
@param_function(start_date="2005-01-01", end_date="2022-12-31")
def rate_abzugsfähige_altersvorsorgeaufwendungen(
    parameter_einführungsfaktor_altersvorsorgeaufwendungen: PiecewisePolynomialParamValue,
    policy_year: int,
    xnp: ModuleType,
) -> dict[str, Any]:
    """Calculate introductory factor for pension expense deductions which depends on the
    current year as follows:

    In the years 2005-2025 the share of deductible contributions increases by
    2 percentage points each year from 60% in 2005 to 100% in 2025.

    Reference: § 10 Abs. 1 Nr. 2 Buchst. a und b EStG


    """
    return piecewise_polynomial(
        x=policy_year,
        parameters=parameter_einführungsfaktor_altersvorsorgeaufwendungen,
        xnp=xnp,
    )
```

hmgaudecker: Ah, sure, sorry! But then why do you need that `param_function` in the first place? Just get rid of the function and do the lookup in the `policy_function` just above.

Issue #24, opened 2025-07-28
BUG: Template returns several variations of same input variable


MImmesberger:
### Bug description

It seems like our templates still don't return the smallest possible set of input data. Don't have the time right now to reproduce this in METTSIM. But doing this in GETTSIM:

```python
TT_TARGETS = {
    "arbeitslosengeld_2": {"betrag_m_bg": "alg2_amount"},  # Monthly ALG II amount per household
}

PROVIDED_AS_INPUT = {
    "p_id": pd.Series([0]),  # Required person IDs
    "sozialversicherung": {"rente": {"altersrente": {"betrag_m": pd.Series([0.0])}}}, # Monthly pension amounts
    "einnahmen": {"renten": {"gesetzliche_m": pd.Series([0.0])}}, # Monthly statutory pension income
    "elterngeld": {"betrag_m": pd.Series([0.0])},  # Monthly parental allowance
}

input_data = InputData.tree(tree=PROVIDED_AS_INPUT)

pruned_inputs_template = main(
    main_target=MainTarget.templates.input_data_dtypes,
    policy_date_str="2025-01-01",
    tt_targets=TTTargets(tree=TT_TARGETS),
    input_data=input_data,
    include_warn_nodes=False,
)
```

returns both `vermögen` and `vermögen_wthh` as required inputs (also, `vermögen_bg` doesn't show up; should keep that in mind when trying to find the bug). Linking the [PR](https://github.com/ttsim-dev/gettsim/issues/1049) that was supposed to handle cases like this.

Also, tagging @JuergenWiemers because he reported the issue!

hmgaudecker:
Fixed in #32

PR #25, opened 2025-07-31
Add ability to create a pruned DAG plot

JuergenWiemers:
When users create an input template with `MainTarget.templates.input_data_dtypes` and provide inputs that would normally be calculated endogenously to the `input_data` argument, TTSIM creates a "pruned" DAG in which the ancestors of the `input_data` variables are dropped if they are not required to calculate the desired targets (essentially the  previous `columns_overriding_functions` functionality). Thus, the returned template returns the minimum set of required inputs (ignoring #24 for a second), for example:

```python
TT_TARGETS = {
    "arbeitslosengeld_2": {"betrag_m_bg": "alg2_benefit"}, 
}
PROVIDED_AS_INPUT = {
    "p_id": pd.Series([0]),
    "einnahmen": {"renten": {"gesetzliche_m": pd.Series([0.0])}},
    "elterngeld": {"betrag_m": pd.Series([0.0])},
}
input_data = InputData.tree(tree=PROVIDED_AS_INPUT)

pruned_inputs_template = main(
    policy_date_str="2025-01-01",
    tt_targets=TTTargets(tree=TT_TARGETS),
    input_data=input_data,
    include_warn_nodes=False,
)
```

However, when plotting ancestors of the same target(s), we can currently only get a plot containing _all_ ancestors, even if some of them are not required for calculation given the user inputs. For example, 

```python
  from __future__ import annotations
  from gettsim import plot
  
  fig = plot.dag.tt(
      policy_date_str="2025-01-01",
      node_selector=plot.dag.NodeSelector(
          type="ancestors",
          node_paths=[("arbeitslosengeld_2", "betrag_m_bg"),]
      ),
      include_params=False,
      show_node_description=True,
  )
```

gives 
test
 (Unfortunately, it seems Github doesn't allow to upload the interactive HTML plot.)

including many nodes from. e.g., the pension calculations that are not required if `einnahmen__rente__gesetzliche_m` is provided as input by the user.

This PR introduces a new field `input_node_paths` (the name is probably not optimal) for the `NodeSelector` type that essentially allows to mimick the pruning possible with `MainTarget.templates.input_data_dtypes`. For example, 

```python
fig = plot.dag.tt(
    policy_date_str="2025-01-01",
    node_selector=plot.dag.NodeSelector(
        type="ancestors",
        node_paths=[("arbeitslosengeld_2", "betrag_m_bg"),],
        input_node_paths=[
            ("einnahmen", "renten", "gesetzliche_m"),
            ("elterngeld", "betrag_m"),
            ("arbeitslosengeld_2", "regelbedarf_m"),
            ("arbeitslosengeld_2", "nettoeinkommen_vor_abzug_freibetrag_m"),
            ("wohngeld", "anspruchshöhe_m_wthh"),
        ],
    ),
    include_params=False,
    show_node_description=True,
    output_path="test_pruned.html",
)
```

creates 
test_pruned

The plot now represents the DAG that is actually used for calculating the target(s) if the variables in `input_node_paths` are provided by the user as inputs. The approach also works if `type="descendants"` is used. In that case, the descendants of the `input_node_paths` are  truncated from the plot.

The implementation is essentially completely ‘vibe-coded’ and should be understood more as a basis for discussion than as an actual PR. I am not even sure if this has any real value for users. My whole motivation for this was "wouldn't it be nice to see a DAG representation that resembles the DAG actually used in the calculations given my inputs?".

### DIFF
--- a/src/ttsim/plot/dag/tt.py
+++ b/src/ttsim/plot/dag/tt.py
@@ -19,6 +19,8 @@ from ttsim.tt import (
 
 if TYPE_CHECKING:
     from pathlib import Path
+    from typing import Any
+    from collections.abc import Callable
 
     import plotly.graph_objects as go
 
@@ -35,6 +37,7 @@ class NodeSelector:
     node_paths: list[tuple[str, ...]]
     type: Literal["neighbors", "descendants", "ancestors", "nodes"]
     order: int | None = None
+    input_node_paths: list[tuple[str, ...]] | None = None
 
 
 @dataclass(frozen=True)
@@ -44,6 +47,7 @@ class _QNameNodeSelector:
     qnames: list[str]
     type: Literal["neighbors", "descendants", "ancestors", "nodes"]
     order: int | None = None
+    input_qnames: list[str] | None = None
 
 
 def tt(
@@ -54,6 +58,7 @@ def tt(
     include_params: bool = True,
     show_node_description: bool = False,
     output_path: Path | None = None,
+    input_node_paths: list[tuple[str, ...]] | None = None,
 ) -> go.Figure:
     """Plot the TT DAG.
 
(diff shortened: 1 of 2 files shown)


MImmesberger:
#32 should fix this once merged

PR #26, opened 2025-07-31
Fix #24

### LINKED ISSUE #24: BUG: Template returns several variations of same input variable

JuergenWiemers:
Closes #24.

The previous first `if` case

```python 
        if (
            base_name not in qname_dtype_tree
            and base_name not in cleaned_qname_dtype_tree
            and base_name in policy_inputs
        ):
```

checks if:

1. The base variable (e.g., `vermögen`) wasn't identified as needed by the DAG ...
2. ... and the base variable wasn't already added to the template...
3. ... and the base variable is a known policy input.

So, if an aggregated variable (e.g., `vermögen_wthh`) was identified as being needed by the DAG, the base name was added to the template instead.

However, this did _not_ catch the case in which both, the base variable and any aggregated variable are required by the DAG, because then `base_name not in qname_dtype_tree == False`.

A test is currently missing because I don't know how to do that for METTSIM...

### DIFF
--- a/src/ttsim/interface_dag_elements/templates.py
+++ b/src/ttsim/interface_dag_elements/templates.py
@@ -77,7 +77,24 @@ def input_data_dtypes(
             continue
 
         match = pattern_all.fullmatch(qn)
-        base_name = match.group("base_name")
+        base_name = match.group("base_name") if match else qn
+
+        # Skip aggregated variables when both a base name and an aggregated
+        # form of the base name is required by the DAG. Fixes #24.
+        if (
+            match
+            and match.group("grouping")
+            and base_name in policy_inputs
+            and qn not in policy_inputs
+        ):
+            # Add the base variable to the template instead of the aggregated one
+            if base_name not in cleaned_qname_dtype_tree:
+                cleaned_qname_dtype_tree[base_name] = scalar_type_to_array_type(
+                    policy_inputs[base_name].data_type
+                )
+            # Skip this aggregated variable - the base variable will be added instead
+            continue
+
         if (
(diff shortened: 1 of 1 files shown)


MImmesberger:
#32 should fix this once merged

PR #32, opened 2025-08-04
Build inputs template and plotting DAG from specialized environment based on `policy_input`s

### LINKED ISSUE #27: ENH: Creating a template when providing input data

(I feel like there is already an issue regarding that, but I couldn't find it)

The output of the dtype templates is a bit awkward when providing input data. Take, for example, this DAG (a stylized version of the wealth tax from METTSIM):

Image

Note that the DAG contains variables that are needed to create the fam_id and kin_id. Everything works as expected when calling 

```python
main(
    policy_date_str="2000-01-01",
    orig_policy_objects={"root": middle_earth.ROOT_PATH},
    tt_targets={"tree": {"wealth_tax": {"amount_y": None}}},
)
```

But when adding some unrelated input data:

```python
import numpy as np
from ttsim.main_args import InputData

main(
    policy_date_str="2000-01-01",
    orig_policy_objects={"root": middle_earth.ROOT_PATH},
    tt_targets={"tree": {"wealth_tax": {"amount_y": None}}},
    input_data=InputData.tree(
        tree={
            "p_id": np.array([0, 1, 2]),
            "payroll_tax": {
                "amount_y": np.array([1000, 0, 0]),
            },
        }
    ),
)
```

the result does not contain the variables needed to compute `kin_id` and `fam_id`. This is no surprise as `wealth_kin` and `wealth_fam` are automatically created functions, which we don't create when input data is provided. This is expected behavior, but confusing when creating the templates.

I didn't look for a solution yet. Maybe we have to do something similar as with the plotting DAG: Select each node of the regular TT DAG as a target and create a second DAG which we use for the template (didn't think this through).

MImmesberger:

Closes #27, #30, #24

This PR tries to fix the root problem of many issues we had with the creation of templates and plotting: derived functions are part of the `specialized_environment` only if they can be build from input data or other `policy_function`s. 

This was an issue because we got incomplete templates or plotting DAGs when using our usual machinery, see #27 for a simple example.

This PR does the following:
-  Add a module `specialized_envrionment_for_plotting_and_templates` that is quite similar to `specialized_environment`. The main difference is that derived functions are created from policy inputs as well. This is useful because for templates and plotting, we want them to be there even if the user doesn't provide the necessary data to actually compute those nodes. Also, we only need qnames of the input data to build this environment.
- Add kwargs to plotting interface
- Add `wealth_tax` to METTSIM to produce the example from #27 
- Add `df_with_nested_columns` as return option for templates

ToDos / Issues:
- [x] Return template as df
- [x] Test kwargs argument to plot_tt and adjust interface of plot_interface_dag
- [ ] Make GETTSIM twin PR

### DIFF
--- a/src/ttsim/interface_dag_elements/specialized_environment_from_policy_inputs.py
+++ b/src/ttsim/interface_dag_elements/specialized_environment_from_policy_inputs.py
@@ -8,11 +8,12 @@ when creating an input data template or plotting it.
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, overload
+from typing import TYPE_CHECKING, Literal, overload
 
 import dags.tree as dt
 from dags import create_dag
 
+from ttsim.interface_dag_elements import specialized_environment
 from ttsim.interface_dag_elements.automatically_added_functions import TIME_UNIT_LABELS
 from ttsim.interface_dag_elements.interface_node_objects import (
     InterfaceFunction,
@@ -36,6 +37,7 @@ from ttsim.tt.column_objects_param_function import (
 from ttsim.tt.param_objects import ParamObject
 
 if TYPE_CHECKING:
+    import datetime
     from collections.abc import Callable
     from types import ModuleType
     from typing import Any
@@ -45,7 +47,10 @@ if TYPE_CHECKING:
     from ttsim.typing import (
         OrderedQNames,
         PolicyEnvironment,
+        QNameData,
         SpecEnvWithoutTreeLogicAndWithDerivedFunctions,
+        SpecEnvWithPartialledParamsAndScalars,
+        SpecEnvWithProcessedParamsAndScalars,
         UnorderedQNames,
     )
 
@@ -107,23 +112,72 @@ def without_tree_logic_and_with_derived_functions(
 
 
 @interface_function()
-def complete_dag(
-    without_tree_logic_and_with_derived_functions: SpecEnvWithoutTreeLogicAndWithDerivedFunctions,
-    tt_targets__qname: OrderedQNames,
-) -> nx.DiGraph:
-    """Create the complete DAG.
-
-    Transforms non-callable nodes into callables to include them as nodes in the DAG.
-    """
-    functions = {
+def without_processed_data_nodes_with_dummy_callables(
+    without_tree_logic_and_with_derived_functions: SpecEnvWithoutTreeLogicAndWithDerivedFunctions,  # noqa: E501
+    processed_data: QNameData,
+) -> SpecEnvWithoutTreeLogicAndWithDerivedFunctions:
+    """Remove nodes that are in the processed data from the policy environment."""
+    return {
         qn: dummy_callable(obj=n, leaf_name=dt.tree_path_from_qname(qn)[-1])
         if not callable(n)
         else n
         for qn, n in without_tree_logic_and_with_derived_functions.items()
+        if qn not in processed_data
     }
+
+
+@interface_function()
+def complete_dag(
+    without_processed_data_nodes_with_dummy_callables: SpecEnvWithoutTreeLogicAndWithDerivedFunctions,  # noqa: E501
+    tt_targets__qname: OrderedQNames,
+) -> nx.DiGraph:
+    """Create the complete DAG.
+
+    The DAG is based on without_tree_logic_and_with_derived_functions because it should
+    include parameters and param_functions. Transforms non-callable nodes into callables
+    to include them as nodes in the DAG.
+    """
     return create_dag(
-        functions=functions,
+        functions=without_processed_data_nodes_with_dummy_callables,
         targets=tt_targets__qname,
+    )
+
+
+@interface_function()
+def with_processed_params_and_scalars(
+    without_tree_logic_and_with_derived_functions: SpecEnvWithoutTreeLogicAndWithDerivedFunctions,  # noqa: E501
+    processed_data: QNameData,
+    backend: Literal["numpy", "jax"],
+    xnp: ModuleType,
+    dnp: ModuleType,
+    evaluation_date: datetime.date | None,
+) -> SpecEnvWithProcessedParamsAndScalars:
+    return specialized_environment.with_processed_params_and_scalars(
+        without_tree_logic_and_with_derived_functions=without_tree_logic_and_with_derived_functions,
+        processed_data=processed_data,
+        backend=backend,
+        xnp=xnp,
+        dnp=dnp,
+        evaluation_date=evaluation_date,
+    )
+
+
+@interface_function()
+def with_partialled_params_and_scalars(
+    with_processed_params_and_scalars: SpecEnvWithProcessedParamsAndScalars,
+    rounding: bool,
+    num_segments: int,
+    backend: Literal["numpy", "jax"],
+    xnp: ModuleType,
+    dnp: ModuleType,
+) -> SpecEnvWithPartialledParamsAndScalars:
+    return specialized_environment.with_partialled_params_and_scalars(
+        with_processed_params_and_scalars=with_processed_params_and_scalars,
+        rounding=rounding,
+        num_segments=num_segments,
+        backend=backend,
+        xnp=xnp,
+        dnp=dnp,
     )
 
 
--- a/src/ttsim/interface_dag_elements/specialized_environment_from_policy_inputs.py
+++ b/src/ttsim/interface_dag_elements/specialized_environment_from_policy_inputs.py
@@ -1,0 +1,136 @@
+"""Builds the specialized environment from policy inputs.
+
+The main difference to `specialized_environment` is that policy inputs are taken as a
+basis for derived functions alongside with the user-provided input data. This is useful
+for applications where users are interested in the DAG itself (not its execution), e.g.
+when creating an input data template or plotting it.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Literal
+from dataclasses import dataclass
+
+import dags.tree as dt
+from dags import create_dag
+
+from ttsim.interface_dag_elements.interface_node_objects import interface_function, interface_input
+from ttsim.interface_dag_elements.shared import dummy_callable
+from ttsim.interface_dag_elements.shared import (
+    get_re_pattern_for_all_time_units_and_groupings,
+)
+from ttsim.interface_dag_elements.automatically_added_functions import TIME_UNIT_LABELS
+from ttsim.interface_dag_elements.specialized_environment import (
+    _remove_tree_logic_from_policy_environment,
+    _add_derived_functions,
+)
+
+
+if TYPE_CHECKING:
+    import networkx as nx
+
+    from ttsim.typing import (
+        OrderedQNames,
+        PolicyEnvironment,
+        SpecEnvWithoutTreeLogicAndWithDerivedFunctions,
+        UnorderedQNames,
+    )
+
+
+@dataclass(frozen=True)
+class QNameNodeSelector:
+    """Select nodes from the DAG."""
+
+    qnames: set[str]
+    type: Literal["neighbors", "descendants", "ancestors", "nodes"]
+    order: int | None = None
+
+
+@interface_input()
+def qname_node_selector() -> QNameNodeSelector:
+    """The qname node selector."""
+
+
+@interface_function()
+def qnames_to_derive_functions_from(
+    labels__input_columns: UnorderedQNames,
+    labels__policy_inputs: UnorderedQNames,
+    labels__grouping_levels: OrderedQNames,
+) -> UnorderedQNames:
+    """The qnames to derive functions from.
+    
+    Derived functions should be created based on the actual input columns, and the
+    policy inputs (if their base name is not already in the input columns).
+    """
+    pattern_all = get_re_pattern_for_all_time_units_and_groupings(
+        time_units=list(TIME_UNIT_LABELS),
+        grouping_levels=labels__grouping_levels,
+    )
+    base_names_input_columns = {
+        pattern_all.fullmatch(qn).group("base_name")
+        for qn in labels__input_columns
+    }
+
+    out = set(labels__input_columns)
+    for pi in labels__policy_inputs:
+        match = pattern_all.fullmatch(pi)
+        base_name = match.group("base_name")
+        if base_name in base_names_input_columns:
+            continue
+        out.add(pi)
+    return out
+
+
+@interface_function()
+def without_tree_logic_and_with_derived_functions(
+    policy_environment: PolicyEnvironment,
+    qnames_to_derive_functions_from: UnorderedQNames,
+    tt_targets__qname: OrderedQNames,
+    labels__top_level_namespace: UnorderedQNames,
+    labels__grouping_levels: OrderedQNames,
+    qname_node_selector: QNameNodeSelector,
+) -> SpecEnvWithoutTreeLogicAndWithDerivedFunctions:
+    """Return a flat policy environment with derived functions.
+
+    Two steps:
+    1. Remove all tree logic from the policy environment.
+    2. Add derived functions to the policy environment.
+
+    """
+    qname_env_without_tree_logic = _remove_tree_logic_from_policy_environment(
+        qname_env=dt.flatten_to_qnames(policy_environment),
+        labels__top_level_namespace=labels__top_level_namespace,
+    )
+    return _add_derived_functions(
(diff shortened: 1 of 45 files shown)


MImmesberger:
@JuergenWiemers This PR should fix all the stuff we talked about in the last couple of days. I hope I didn't miss any corner case. If you find the time, feel free to try it out and complain if I missed something!

src/ttsim/plot/dag/tt.py
```diff
@@ -202,23 +223,13 @@ def _get_tt_dag_with_node_metadata(
...
-        qn: dummy_callable(obj=n, leaf_name=dt.tree_path_from_qname(qn)[-1])
-        if not callable(n)
-        else n
-        for qn, n in env.items()
-    }
-
-
 def _get_node_descriptions(
-    env: SpecEnvWithoutTreeLogicAndWithDerivedFunctions,
+    specialized_environment: PolicyEnvironment,
```
hmgaudecker, line 227: This is very confusing naming (=namespace, but PolicyEnvironment annotation...). I have no idea what this should be.

hmgaudecker:
Thanks! I looked through most of it but don't quite get the exact workings yet.

In particular, could you explain in the docstrings or here how the issue we had with respect to providing some input data and requesting all targets is resolved in `specialized_envrionment_from_policy_inputs`?

MImmesberger:
> Oh, forgot to voice the feeling I had that maybe we can simplify more in specialized_environment now? That is, didn't we include what we factor out now as a special case?

I don't think we have to adjust `specialized_environment`, we handles this special case in `main.py` and `labels.py`. But you're right, I forgot to remove those!

src/ttsim/plot/dag/tt.py
```diff
@@ -113,45 +120,61 @@ def tt(
...
             complete_dag=complete_dag,
             node_selector=node_selector,
         )
 
     if not include_params:
         selected_dag.remove_nodes_from(
-            [qn for qn, v in env.items() if isinstance(v, (ParamObject, ParamFunction))]
+            [
+                qn
+                for qn, v in policy_environment.items()
```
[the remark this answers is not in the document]
MImmesberger: Ah you're right our tests didn't cover nested policy envs. Also, filtering out time-converted params didn't work. Added this now.

src/ttsim/interface_dag_elements/specialized_environment_from_policy_inputs.py
```diff
@@ -0,0 +1,224 @@
...
+        tt_targets=labels__column_targets + labels__param_targets,
+        input_columns=qnames_to_derive_functions_from,
+        grouping_levels=labels__grouping_levels,
+    )
+
+
+@interface_function()
+def without_processed_data_nodes_with_dummy_callables(
+    without_tree_logic_and_with_derived_functions: SpecEnvWithoutTreeLogicAndWithDerivedFunctions,  # noqa: E501
+    processed_data: QNameData,
```
hmgaudecker, line 118: I think this module should only ever depend on `input_data__columns`. If calling the version from `specialized_environment`, just create a dummy data structure from that?

MImmesberger:
> Shouldn't it also show housing_benefits__income__amount_m_fam?

No, because `housing_benefits__income__amount_m_fam` is a derived function (not what I would consider hard coded). One could throw a different warning there of course.

> I cannot set include_warn_nodes=False with ttsim.plot.dag.tt. I guess the idea is to move the plotting functionality to something like main(main_target=MainTarget.plot.dag.tt, ...) at a later point?

Good point, that should be `False` by default in that case.

We want to keep the plotting infrastructure outside of `main`. Allows for a bit more flexibility when providing plotly kwargs (not implemented yet, but the idea is that users can customize the plot as they please) and it would bloat the interface DAG.

> It feels a bit awkward to have to provide a "dummy dataset" as input_data just to get a pruned graph. I mean, the data don't really matter, what matters is just the nodes, right?

I agree. I thought it might be more convenient for users to pass input data using the same syntax as with `main`, but it does more harm than good here, I believe.

src/ttsim/interface_dag_elements/specialized_environment_from_policy_inputs.py
```diff
@@ -58,49 +57,41 @@
...
-    When computing nodes from this namespace (helpers to create templates or plot the
-    DAG), we don't necessarily need to have processed data, as processed data only
-    customizes the targets.
-    """
-    return processed_data
+def input_columns_from_input_data(
+    labels__input_columns: UnorderedQNames,
+) -> UnorderedQNames:
+    """The (qualified) column names in the input data."""
+    return labels__input_columns
```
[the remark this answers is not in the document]
MImmesberger: I agree. I tried to make it fool-proof (makes it very clear that `input_columns_is_empty_set` should never be part of the DAG outside of templates or DAG plot creation. But we have good reviewer(s), so no need to think too much about errors of future devs 😄

src/ttsim/interface_dag_elements/specialized_environment_from_policy_inputs.py
```diff
@@ -0,0 +1,229 @@
...
+def with_processed_params_and_scalars(
+    without_tree_logic_and_with_derived_functions: SpecEnvWithoutTreeLogicAndWithDerivedFunctions,  # noqa: E501
+    backend: Literal["numpy", "jax"],
+    xnp: ModuleType,
+    dnp: ModuleType,
+    evaluation_date: datetime.date | None,
+) -> SpecEnvWithProcessedParamsAndScalars:
+    return specialized_environment.with_processed_params_and_scalars(
+        without_tree_logic_and_with_derived_functions=without_tree_logic_and_with_derived_functions,
+        processed_data={},
```
hmgaudecker, line 165: I am not sure this should work because in `specialized_environment.with_processed_params_and_scalars`, we are removing the contents of `processed_data` from the graph. This will leave all nodes in there, which is probably behind the failure. Hence my comment on creating some fake processed data.

src/ttsim/plot/dag/tt.py
```diff
@@ -41,70 +46,105 @@ class NodeSelector:
...
 
-    qnames: list[str]
+    qnames: set[str]
     type: Literal["neighbors", "descendants", "ancestors", "nodes"]
     order: int | None = None
 
 
 def tt(
-    policy_date_str: str,
+    # Args specific to TTSIM plotting
```
hmgaudecker, line 55:
```suggestion
    *,
    # Args specific to TTSIM plotting
```
Make the function keyword-only.

MImmesberger:
I think I finally have a solution that works. The basic problem is that we 

- have to select all available nodes as targets for the DAG in order to allow for plotting of neighbors and descendants (in other words, given the node the user is interested in, we don't know ex-ante which nodes should be "targets" in `dags` terminology)
- but when adding input data, we want to get rid of sets of nodes overridden by one input column

The solution is to first create a DAG setting all nodes as targets (but removing some functions because of data inputs) and then removing all nodes that are not ancestors or descendants of any of the targets.

This comes a bit out of the blue, so here is an example:

Screenshot 2025-08-07 at 19 29 20

- The user is interested in "c". We need to accommodate all possible types of plots: i) a plot of the entire DAG, ii) ancestors of "c", iii) descendents of "c", iv) neighbors of "c".
- If no input data is given, its easy: Create the entire DAG and then choose the nodes needed.
- If "b" is overwritten by input data, we are left with this DAG:

Screenshot 2025-08-07 at 19 30 50

- Because we want to get rid of "d" in the final output, we eliminate it by searching for the ancestors /descendants of "c". Then, we can, as in the case with no input data, just choose the nodes we need for our plot.

Apart from that, I removed the `NodeSelector` dataclass. This way, users can set targets as they do when computing taxes and transfers, no need to introduce a new concept.

src/ttsim/main_target.py
```diff
@@ -112,11 +123,12 @@ class Targets(MainTargetABC):
 
 @dataclass(frozen=True)
 class Labels(MainTargetABC):
+    all_qnames_in_policy_environment: str = "labels__all_qnames_in_policy_environment"
```
hmgaudecker, line 126: Please move this and `policy_inputs` down in the list, these are among the least important / purely internal ones. Instead, `input_columns` should move up now that it might actually be used.

src/ttsim/plot/dag/tt.py
```diff
@@ -202,23 +258,13 @@ def _get_tt_dag_with_node_metadata(
...
-
 def _get_node_descriptions(
-    env: SpecEnvWithoutTreeLogicAndWithDerivedFunctions,
+    without_tree_logic_and_with_derived_functions: SpecEnvWithoutTreeLogicAndWithDerivedFunctions,  # noqa: E501
 ) -> dict[str, str]:
     """Get the descriptions of the nodes in the environment."""
+    qn_env = dt.flatten_to_qnames(without_tree_logic_and_with_derived_functions)
     out = {}
-    for qn, n in env.items():
+    for p, n in qn_env.items():
```
hmgaudecker, line 267: Why change the running variable? I'd go for `qn, node` actually -- ended up being a sufficiently long loop.

src/ttsim/plot/dag/tt.py
```diff
@@ -229,49 +275,53 @@ def _get_node_descriptions(
...
+    selection_type: Literal["neighbors", "descendants", "ancestors", "nodes"],
+    selection_depth: int | None = None,
 ) -> nx.DiGraph:
     """Select nodes based on the node selector."""
     selected_nodes: set[str] = set()
-    if node_selector.type == "nodes":
-        selected_nodes.update(node_selector.qnames)
-    elif node_selector.type == "ancestors":
-        for node in node_selector.qnames:
+    if selection_type == "nodes":
```
hmgaudecker, line 294: Please use the same order as in the `Literal[]` definition, makes it easier to read.

src/ttsim/plot/dag/tt.py
```diff
@@ -229,49 +275,53 @@ def _get_node_descriptions(
...
+            f"Invalid selection type: {selection_type}. "
             "Choose one of 'nodes', 'ancestors', 'descendants', or 'neighbors'."
         )
         raise ValueError(msg)
 
-    dag_copy = copy.deepcopy(complete_dag)
-    dag_copy.remove_nodes_from(set(complete_dag.nodes) - set(selected_nodes))
+    dag_copy = copy.deepcopy(complete_tt_dag)
+    dag_copy.remove_nodes_from(set(complete_tt_dag.nodes) - set(selected_nodes))
     return dag_copy
```
hmgaudecker, line 325:
```suggestion
    return copy.deepcopy(complete_tt_dag).remove_nodes_from(set(complete_tt_dag.nodes) - set(selected_nodes))
```
If you find that to be too much for one line, create:

```py
unnecessary_nodes = set(complete_tt_dag.nodes) - set(selected_nodes)
```

upfront. Makes the logic clear instead of implementation details.

src/ttsim/plot/dag/tt.py
```diff
@@ -113,46 +135,80 @@ def tt(
...
+            elif (
+                isinstance(v, TimeConversionFunction)
+                and hasattr(v, "source")
+                and v.source in without_tree_logic_and_with_derived_functions
+                and isinstance(
+                    without_tree_logic_and_with_derived_functions[v.source],
+                    (ParamObject, ParamFunction),
+                )
+            ):
+                # Also add time-converted params
```
hmgaudecker, line 208:
```suggestion
                # Also schedule time-converted params for removal.
```
We add to the set, but the goal is different...

src/ttsim/plot/dag/tt.py
```diff
@@ -229,49 +275,53 @@ def _get_node_descriptions(
...
-    complete_dag: nx.DiGraph,
-    node_selector: _QNameNodeSelector,
+def select_nodes_from_dag(
+    complete_tt_dag: nx.DiGraph,
+    tt_targets_qnames: set[str],
+    selection_type: Literal["neighbors", "descendants", "ancestors", "nodes"],
+    selection_depth: int | None = None,
 ) -> nx.DiGraph:
     """Select nodes based on the node selector."""
     selected_nodes: set[str] = set()
```
hmgaudecker, line 293: Can you remove this and assign directly in each of the four cases? Should be a simple `set()` call for nodes and set comprehensions for the others.

src/ttsim/plot/dag/tt.py
```diff
@@ -2,109 +2,131 @@
...
-    node_selector
-        The node selector. Default is None, i.e. the entire DAG is plotted.
-    title
-        The title of the plot.
+    selection_type
+        The type of the DAG to plot. Can be one of:
+        - "neighbors": Plot the neighbors of the target nodes.
+        - "descendants": Plot the descendants of the target nodes.
+        - "ancestors": Plot the ancestors of the target nodes.
+        - "nodes": Plot the target nodes.
```
hmgaudecker, line 70: You mentioned it in your edit, but I truly have no idea what is supposed to happen for "nodes".

hmgaudecker (commented):
Many thanks! I think the background stuff is great now!

I have not tried the plotting myself, leaving that to @JuergenWiemers to check! Also, please open a parallel PR in GETTSIM to see whether everything works as expected there or we might be missing something.

> Apart from that, I removed the `NodeSelector` dataclass. This way, users can set targets as they do when computing taxes and transfers, no need to introduce a new concept.

Makes sense, though I am not perfectly sure from looking at the example etc. what to expect when `tt_targets` is `c` and I ask for `descendants`. Should `e` ever show up? Maybe that is just confusion on my side because of the overloading of the term `target`, but I don't think `tt_targets` and `node_selectors` (leaving the old one for lack of a better word) can be integrated into one concept.

src/ttsim/plot/dag/tt.py
MImmesberger: `remove_nodes_from` is an inplace method which is why I do it this way. Returning and
removing nodes in the same line won't work I believe

hmgaudecker: My bad, sorry for the noise!

src/ttsim/plot/dag/tt.py
```diff
@@ -2,109 +2,134 @@
...
+            - "descendants": Plot the descendants of the primary nodes.
+            - "ancestors": Plot the ancestors of the primary nodes.
+            - "nodes": Plot the primary nodes only.
+        If not provided, the entire DAG is plotted.
+    selection_depth
+        The depth of the selection. Only used if selection_type is "neighbors",
+        "descendants", or "ancestors".
     include_params
-        Include param functions when plotting the DAG.
+        Include param functions when plotting the DAG. Default is True.
```
hmgaudecker, line 81:
```suggestion
        Include params and param functions when plotting the DAG. Default is True.
```
?

src/ttsim/main_target.py
```diff
@@ -113,13 +124,14 @@ class Targets(MainTargetABC):
 @dataclass(frozen=True)
 class Labels(MainTargetABC):
     column_targets: str = "labels__column_targets"
-    grouping_levels: str = "labels__grouping_levels"
+    input_columns: str = "labels__input_columns"
```
hmgaudecker, line 127: Apologies for the micro-management, but this needs to go up one more line because the three types of targets have to be one block for my sanity...

hmgaudecker (approved):
Very nice, thanks! Feel free to merge if @JuergenWiemers is happy with the behaviour and GETTSIM does not show anything unexpected!

MImmesberger:
Thanks for trying it out! @JuergenWiemers 

> I noticed just a tiny asymmetry in the DAG plots

I agree with HM here. The DAG plot is supposed to show what GETTSIM calculates eventually. Including the input node in the plot when we use `selection_type='ancestors'` makes sense because the input node is a root node of the DAG.

With `selection_type='descendants'`, the input node is completely independent of its ancestors; it would be confusing to indicate a relationship between them by including it in the DAG.

hmgaudecker:
Thanks for the explanation! 

> I thought the idea of the descendants plot with additional "known nodes" was to prune the immediate descendants of the known nodes, but the known nodes itself should be included. This is what I mean by "asymmetry in the plots".

The behaviour could be changed in this way, ofc, but I do not think it would be helpful in a broader set of use cases. Think about GETTSIM and you are simply throwing your input data at the plotting function _(the column names of the data end up being `labels__input_columns`)_. If you asked for the descendants of a single node, you would still get **all** other input columns as orphaned nodes.

Maybe a different way to view it: There is no way to recur to the structure of the graph that would obtain if you had not passed this particular input data. _Well, maybe we could achieve that, but I am pretty sure that would be like opening Pandora's box..._

In order to get the plot you want, you could just pass `orc_hunting_bounty__amount` as a primary node, too.

PR #34, opened 2025-08-06
Optimize JAX performance in data preparation pipeline

JuergenWiemers:
Let me start by letting the numbers speak for themselves:

```python
Benchmark Comparison Report
Generated: 2025-08-06T15:13:33.129062
Main branch file: benchmark_results_20250806_150436.json
PR branch file: benchmark_results_20250806_150643.json

==========================================================================================
JAX BACKEND COMPARISON: Main Branch vs PR Branch
==========================================================================================
Households  JAX hash main  JAX hash PR    JAX main (s)   JAX PR (s)     Speedup     
------------------------------------------------------------------------------------------
32,767      9a97a0ef       9a97a0ef       2.4546         0.7370         3.33x       
32,768      c0a7c7e9       c0a7c7e9       2.0527         0.5846         3.51x       
65,536      f873c7c8       f873c7c8       3.4010         0.6154         5.53x       
131,072     84608262       84608262       6.2885         0.7180         8.76x       
262,144     2fb136ba       2fb136ba       12.1383        1.0488         11.57x      
524,288     dd61a524       dd61a524       23.6647        1.9196         12.33x      
1,048,576   6ed4af0a       6ed4af0a       46.8753        3.8041         12.32x      
2,097,152   ff9f600f       ff9f600f       96.5656        8.0620         11.98x      

==========================================================================================
NUMPY BACKEND COMPARISON: Main Branch vs PR Branch
==========================================================================================
Households  NP hash main   NP hash PR     NP main (s)    NP PR (s)      Speedup     
------------------------------------------------------------------------------------------
32,767      2584ff76       2584ff76       0.1028         0.0907         1.13x       
32,768      a4ec6c3f       a4ec6c3f       0.1031         0.0692         1.49x       
65,536      e18ee207       e18ee207       0.1767         0.1279         1.38x       
131,072     e949fd41       e949fd41       0.4076         0.2277         1.79x       
262,144     a8fe5730       a8fe5730       0.7846         0.4868         1.61x       
524,288     c7c95469       c7c95469       1.3177         0.9271         1.42x       
1,048,576   1fbbd4ff       1fbbd4ff       2.8963         2.1271         1.36x       
2,097,152   81209a75       81209a75       6.0628         4.9379         1.23x       

======================================================================
SUMMARY STATISTICS
======================================================================

NUMPY Backend:
--------------------
  Average speedup: 1.43x
  Maximum speedup: 1.79x
  Minimum speedup: 1.13x
  Successful runs: 8/8
  Hash mismatches: 0/8
  ✓ All results are numerically identical

JAX Backend:
--------------------
  Average speedup: 8.67x
  Maximum speedup: 12.33x
  Minimum speedup: 3.33x
  Successful runs: 8/8
  Hash mismatches: 0/8
  ✓ All results are numerically identical
```

### Explanation/Remarks: 

- Profiling TTSIM (`main(main_targets=MainTarget.processed_data,...)`) for large datasets with the JAX backend shows that the majority of the runtime is spent in `data_converers.df_with_mapped_columns_to_flat_data` and `fail_if.input_data_is_invalid`
  
  gettsim_jax_1_000_000_main

- Regarding `data_converers.df_with_mapped_columns_to_flat_data`: For the JAX backend, it is much faster to create NumPy arrays first and then convert them to JAX arrays rather than to immediately create JAX arrays.
- Regarding `fail_if.input_data_is_invalid`: The loop over the JAX array is very slow. I think it should be safe to convert `p_id` to a NumPy array and then use Pandas for the duplicates detection.
- I expected the NumPy backend to be more or less unaffected, but it got a little faster as well...
- In case you are wondering "Why 32,767 and 32,768 households?" I found out by experimenting that `xnp.asarray(numpy_array, copy=False)` still copies for arrays smaller than 1MB; at exactly 1MB, JAX gives up and stops creating copies. The duplicated example household has four members, thus 32,768 * 4 * 8 = 1,048,567 = 1 MB (for a 64 bit int or float column). In the results you can see that 32.767 is (consistently, I checked) slower than 32.768 probably because in the former case arrays still get copied despite setting `xnp.asarray(numpy_array, copy=False)`.
- 🔪 ☠️ : You can use the `copy=False` option to mutate "immutable" JAX arrays by mutating the NumPy array that shares the same memory (if the array is >=1MB)! 😮There was probably a heated debate about this "feature" among the JAX developers...
- After the two optimizations the profile looks like this. The bottlenecks are gone.
  
  gettsim_jax_1_000_000_PR

### TODOs

- [x] I am not really sure if this requires new tests. The code just replaces existing code.
- [x] I would like to know if this also results in better performance for JAX-CUDA and/or Metal, but I don't have the hardware to check this...

### DIFF
--- a/src/ttsim/interface_dag_elements/data_converters.py
+++ b/src/ttsim/interface_dag_elements/data_converters.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Any
 
 import dags.tree as dt
+import numpy as np
 import pandas as pd
 
 if TYPE_CHECKING:
@@ -131,12 +132,19 @@ def df_with_mapped_columns_to_flat_data(
 
 
     """
+    # Always use numpy for initial data preparation (performance optimization for JAX)
     path_to_array = {}
     for path, mapper_value in dt.flatten_to_tree_paths(mapper).items():
-        if xnp.isscalar(mapper_value) and not isinstance(mapper_value, str):
-            path_to_array[path] = xnp.asarray([mapper_value] * len(df))
+        if np.isscalar(mapper_value) and not isinstance(mapper_value, str):
+            numpy_array = np.asarray([mapper_value] * len(df))
         else:
-            path_to_array[path] = xnp.asarray(df[mapper_value])
+            numpy_array = np.asarray(df[mapper_value])
+
+        # Convert to target backend array with zero-copy if JAX
+        if xnp.__name__ == "jax.numpy":
+            path_to_array[path] = xnp.asarray(numpy_array, copy=False)
+        else:
+            path_to_array[path] = numpy_array
 
     return path_to_array
 
@@ -164,10 +172,19 @@ def df_with_nested_columns_to_flat_data(
         >>> result
         {("a", "b"): np.array([1, 2, 3]), ("c",): np.array([4, 5, 6])}
     """
-    return {
-        _remove_nan_from_keys(key): xnp.asarray(value)
-        for key, value in df.to_dict(orient="list").items()
-    }
+    # Always use numpy for initial data preparation (performance optimization for JAX)
+    result = {}
+    for key, value in df.to_dict(orient="list").items():
+        clean_key = _remove_nan_from_keys(key)
+        numpy_array = np.asarray(value)
+
+        # Convert to target backend array with zero-copy if JAX
+        if xnp.__name__ == "jax.numpy":
+            result[clean_key] = xnp.asarray(numpy_array, copy=False)
+        else:
+            result[clean_key] = numpy_array
+
+    return result
 
 
 def _remove_nan_from_keys(path: tuple[str | Any, ...]) -> tuple[str, ...]:
--- a/src/ttsim/interface_dag_elements/data_converters.py
+++ b/src/ttsim/interface_dag_elements/data_converters.py
@@ -133,17 +133,18 @@ def df_with_mapped_columns_to_flat_data(
 
 
     """
-    # Always use numpy for initial data preparation (performance optimization for JAX)
     path_to_array = {}
     for path, mapper_value in dt.flatten_to_tree_paths(mapper).items():
+        # Use numpy for array creation if JAX backend is chosen
+        # Performance optimization for JAX, PR #34
         if numpy.isscalar(mapper_value) and not isinstance(mapper_value, str):
             numpy_array = numpy.asarray([mapper_value] * len(df))
         else:
             numpy_array = numpy.asarray(df[mapper_value])
 
-        # Convert to target backend array with zero-copy if JAX
+        # Convert numpy array back to JAX array if JAX backend is chosen
         if backend == "jax":
-            path_to_array[path] = xnp.asarray(numpy_array, copy=False)
+            path_to_array[path] = xnp.asarray(numpy_array)
         else:
             path_to_array[path] = numpy_array
 
@@ -174,17 +175,17 @@ def df_with_nested_columns_to_flat_data(
         >>> result
         {("a", "b"): np.array([1, 2, 3]), ("c",): np.array([4, 5, 6])}
     """
-    # Always use numpy for initial data preparation (performance optimization for JAX)
     result = {}
     for key, value in df.to_dict(orient="list").items():
         clean_key = _remove_nan_from_keys(key)
-        numpy_array = numpy.asarray(value)
 
-        # Convert to target backend array with zero-copy if JAX
+        # Use numpy for array creation if JAX backend is chosen and
+        # immediately convert back to JAX array.
+        # Performance optimization for JAX, PR #34
         if backend == "jax":
-            result[clean_key] = xnp.asarray(numpy_array, copy=False)
+            result[clean_key] = xnp.asarray(numpy.asarray(value))
         else:
-            result[clean_key] = numpy_array
+            result[clean_key] = numpy.asarray(value)
 
     return result
 
--- a/src/ttsim/interface_dag_elements/data_converters.py
+++ b/src/ttsim/interface_dag_elements/data_converters.py
@@ -1,9 +1,9 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Literal
 
 import dags.tree as dt
-import numpy as np
+import numpy
 import pandas as pd
 
 if TYPE_CHECKING:
@@ -84,6 +84,7 @@ def nested_data_to_df_with_mapped_columns(
 def df_with_mapped_columns_to_flat_data(
     mapper: NestedInputsMapper,
     df: pd.DataFrame,
+    backend: Literal["numpy", "jax"],
     xnp: ModuleType,
 ) -> FlatData:
     """Transform a pandas DataFrame to a flattened data structure.
@@ -135,13 +136,13 @@ def df_with_mapped_columns_to_flat_data(
     # Always use numpy for initial data preparation (performance optimization for JAX)
     path_to_array = {}
     for path, mapper_value in dt.flatten_to_tree_paths(mapper).items():
-        if np.isscalar(mapper_value) and not isinstance(mapper_value, str):
-            numpy_array = np.asarray([mapper_value] * len(df))
+        if numpy.isscalar(mapper_value) and not isinstance(mapper_value, str):
+            numpy_array = numpy.asarray([mapper_value] * len(df))
         else:
-            numpy_array = np.asarray(df[mapper_value])
+            numpy_array = numpy.asarray(df[mapper_value])
 
         # Convert to target backend array with zero-copy if JAX
-        if xnp.__name__ == "jax.numpy":
+        if backend == "jax":
             path_to_array[path] = xnp.asarray(numpy_array, copy=False)
         else:
             path_to_array[path] = numpy_array
@@ -151,6 +152,7 @@ def df_with_mapped_columns_to_flat_data(
 
 def df_with_nested_columns_to_flat_data(
     df: pd.DataFrame,
+    backend: Literal["numpy", "jax"],
     xnp: ModuleType,
 ) -> FlatData:
     """Convert a DataFrame with nested columns to a flattened data structure.
@@ -176,10 +178,10 @@ def df_with_nested_columns_to_flat_data(
     result = {}
     for key, value in df.to_dict(orient="list").items():
         clean_key = _remove_nan_from_keys(key)
-        numpy_array = np.asarray(value)
+        numpy_array = numpy.asarray(value)
 
         # Convert to target backend array with zero-copy if JAX
-        if xnp.__name__ == "jax.numpy":
+        if backend == "jax":
             result[clean_key] = xnp.asarray(numpy_array, copy=False)
         else:
             result[clean_key] = numpy_array
--- a/src/ttsim/interface_dag_elements/fail_if.py
+++ b/src/ttsim/interface_dag_elements/fail_if.py
@@ -307,7 +307,7 @@ def input_data_is_invalid(input_data__flat: FlatData, xnp: ModuleType) -> None:
     if p_id.shape[0] == 1:  # Special case for single-row data
         return
 
-    p_id_sorted = xnp.sort(p_id)
+    p_id_sorted = xnp.sort(xnp.asarray(p_id))
     duplicates = xnp.diff(p_id_sorted, append=0) == 0
     if xnp.sum(duplicates) >= 1:
         message = (
--- a/src/ttsim/interface_dag_elements/fail_if.py
+++ b/src/ttsim/interface_dag_elements/fail_if.py
@@ -305,10 +305,9 @@ def input_data_is_invalid(input_data__flat: FlatData) -> None:
 
     # Check for non-unique p_ids
     # Convert to numpy array for performance (avoid slow JAX array iteration)
-    p_id_numpy = numpy.asarray(p_id)
 
     # Use pandas duplicated for efficient duplicate detection
-    p_id_series = pd.Series(p_id_numpy)
+    p_id_series = pd.Series(numpy.asarray(p_id))
     duplicated_mask = p_id_series.duplicated(keep=False)
     non_unique_p_ids = p_id_series[duplicated_mask].unique().tolist()
 
--- a/src/ttsim/interface_dag_elements/fail_if.py
+++ b/src/ttsim/interface_dag_elements/fail_if.py
@@ -283,7 +283,7 @@ def input_data_tree_is_invalid(
 
 
 @fail_function(include_if_any_element_present=["input_data__flat"])
-def input_data_is_invalid(input_data__flat: FlatData) -> None:
+def input_data_is_invalid(input_data__flat: FlatData, xnp: ModuleType) -> None:
     """Fail if the input data is invalid.
 
     Fails if:
@@ -304,17 +304,15 @@ def input_data_is_invalid(input_data__flat: FlatData) -> None:
         raise ValueError(msg)
 
     # Check for non-unique p_ids
-    # Convert to numpy array for performance (avoid slow JAX array iteration)
+    if p_id.shape[0] == 1:  # Special case for single-row data
+        return
 
-    # Use pandas duplicated for efficient duplicate detection
-    p_id_series = pd.Series(numpy.asarray(p_id))
-    duplicated_mask = p_id_series.duplicated(keep=False)
-    non_unique_p_ids = p_id_series[duplicated_mask].unique().tolist()
-
-    if non_unique_p_ids:
+    p_id_sorted = xnp.sort(p_id)
+    duplicates = xnp.diff(p_id_sorted, append=0) == 0
+    if xnp.sum(duplicates) >= 1:
         message = (
             "The following `p_id`s are not unique in the input data:\n\n"
-            f"{non_unique_p_ids}\n\n"
+            f"{p_id_sorted[duplicates]}\n\n"
         )
         raise ValueError(message)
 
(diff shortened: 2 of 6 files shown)


src/ttsim/interface_dag_elements/data_converters.py
```diff
@@ -3,6 +3,7 @@
 from typing import TYPE_CHECKING, Any
 
 import dags.tree as dt
+import numpy as np
```
hmgaudecker, line 6:
```suggestion
import numpy
```

This is the convention we use in GETTSIM in order to be very clear what we are using.

src/ttsim/interface_dag_elements/data_converters.py
```diff
@@ -131,12 +132,19 @@ def df_with_mapped_columns_to_flat_data(
...
-        if xnp.isscalar(mapper_value) and not isinstance(mapper_value, str):
-            path_to_array[path] = xnp.asarray([mapper_value] * len(df))
+        if np.isscalar(mapper_value) and not isinstance(mapper_value, str):
+            numpy_array = np.asarray([mapper_value] * len(df))
         else:
-            path_to_array[path] = xnp.asarray(df[mapper_value])
+            numpy_array = np.asarray(df[mapper_value])
+
+        # Convert to target backend array with zero-copy if JAX
+        if xnp.__name__ == "jax.numpy":
```
hmgaudecker, line 144:
```suggestion
        if backend == "jax":
```
Just add `backend` as an argument to the function.

src/ttsim/interface_dag_elements/data_converters.py
```diff
@@ -164,10 +172,19 @@ def df_with_nested_columns_to_flat_data(
...
     """
-    return {
-        _remove_nan_from_keys(key): xnp.asarray(value)
-        for key, value in df.to_dict(orient="list").items()
-    }
+    # Always use numpy for initial data preparation (performance optimization for JAX)
+    result = {}
+    for key, value in df.to_dict(orient="list").items():
+        clean_key = _remove_nan_from_keys(key)
+        numpy_array = np.asarray(value)
```
hmgaudecker, line 179:
```suggestion
```
Better do below.

src/ttsim/interface_dag_elements/data_converters.py
```diff
@@ -164,10 +172,19 @@ def df_with_nested_columns_to_flat_data(
...
+
+        # Convert to target backend array with zero-copy if JAX
+        if xnp.__name__ == "jax.numpy":
```
hmgaudecker, line 182:
```suggestion
        if backend == "jax":
```
As above

src/ttsim/interface_dag_elements/data_converters.py
```diff
@@ -164,10 +172,19 @@ def df_with_nested_columns_to_flat_data(
...
+            result[clean_key] = xnp.asarray(numpy_array, copy=False)
```
hmgaudecker, line 183:
```suggestion
            result[clean_key] = xnp.asarray(numpy.asarray(value)
, copy=False)
```
Here I am not sure whether the `copy=False` works when we are using the GPU backend. @mj023, could you try that out?

src/ttsim/interface_dag_elements/fail_if.py
```diff
@@ -304,15 +304,13 @@ def input_data_is_invalid(input_data__flat: FlatData) -> None:
...
     # Check for non-unique p_ids
-    p_id_counts: dict[int, int] = {}
-    # Need the map because Jax loop items are 1-element arrays.
-    for i in map(int, p_id):
-        if i in p_id_counts:
-            p_id_counts[i] += 1
-        else:
-            p_id_counts[i] = 1
+    # Convert to numpy array for performance (avoid slow JAX array iteration)
+    p_id_numpy = numpy.asarray(p_id)
```
hmgaudecker, line 308:
```suggestion
```
No need for this assignment.

src/ttsim/interface_dag_elements/fail_if.py
```diff
@@ -304,15 +304,13 @@ def input_data_is_invalid(input_data__flat: FlatData) -> None:
...
 
-    non_unique_p_ids = [i for i, count in p_id_counts.items() if count > 1]
+    # Use pandas duplicated for efficient duplicate detection
+    p_id_series = pd.Series(p_id_numpy)
```
hmgaudecker, line 311:
```suggestion
    p_id_series = pd.Series(numpy.asarray(p_id))
```

src/ttsim/interface_dag_elements/fail_if.py
hmgaudecker, line 1: In general, I am sympathetic to using efficient pandas implementations here instead of cooking up our own thing. Might be useful to think about that in more general context, though -- we have plenty of open issues in GETTSIM for checking data sanity. @MImmesberger, could you make a note on that or create some meta-issue?

hmgaudecker (approved):
Thanks a lot, very nice!!!

I didn't realise yesterday how quick it would be to look at this -- looks great to me, just some stylistic things. 

Approval conditional on this working with the GPU backend, cannot test that myself.

src/ttsim/interface_dag_elements/data_converters.py
JuergenWiemers: The [Jax documentation](https://docs.jax.dev/en/latest/_autosummary/jax.numpy.asarray.html) states:

> - copy ([bool](https://docs.python.org/3/library/functions.html#bool) | None) – optional boolean specifying the copy mode. If True, then always return a copy. If False, then error if a copy is necessary. Default is None, which will only copy when necessary.

If you take that literally, this seems to suggest that `xnp.asarray(numpy_array, copy=False)` will error with the JAX-GPU backend because copying of the NumPy array to the GPU is definitely necessary. However, I find it quite likely that JAX will simply ignore the parameter when CPU->GPU transfer is necessary. You cannot _not_ copy in that case! But yes, impossible to know for sure without trying it out...

mj023: You are right, it does not work with the GPU backend, I just tested it. But that is not a problem, it still works the same, if you just remove `copy=False`. The speedup seems to come from first converting to numpy arrays, this [here](https://github.com/jax-ml/jax/issues/10662) might explain it.

mj023:
Old Code on GPU:

```
============================================================
PERFORMANCE COMPARISON
============================================================
Households  NumPy (s)   JAX (s)     NumPy Hash  JAX Hash    Speedup     
------------------------------------------------------------------------
32,767      0.0514      6.1766      2584ff76    9a97a0ef    1/120.22x   
32,768      0.0377      4.6372      a4ec6c3f    c0a7c7e9    1/123.04x   
65,536      0.0696      8.2590      e18ee207    f873c7c8    1/118.66x   
131,072     0.1301      16.5376     e949fd41    84608262    1/127.12x   
262,144     0.2605      30.7391     a8fe5730    2fb136ba    1/118.01x   
524,288     0.5144      58.4322     c7c95469    dd61a524    1/113.59x   
1,048,576   1.0611      117.1636    1fbbd4ff    6ed4af0a    1/110.42x   
2,097,152   2.1319      234.7411    81209a75    ff9f600f    1/110.11x 
```

New Code on GPU:
```
============================================================
PERFORMANCE COMPARISON
============================================================
Households  NumPy (s)   JAX (s)     NumPy Hash  JAX Hash    Speedup     
------------------------------------------------------------------------
32,767      0.0305      1.1938      2584ff76    9a97a0ef    1/39.20x    
32,768      0.0225      0.6582      a4ec6c3f    c0a7c7e9    1/29.29x    
65,536      0.0423      0.6841      e18ee207    f873c7c8    1/16.17x    
131,072     0.0790      0.7263      e949fd41    84608262    1/9.20x     
262,144     0.1510      0.7738      a8fe5730    2fb136ba    1/5.12x     
524,288     0.3497      0.8522      c7c95469    dd61a524    1/2.44x     
1,048,576   0.7123      0.9904      1fbbd4ff    6ed4af0a    1/1.39x     
2,097,152   1.3929      1.2859      81209a75    ff9f600f    1.08x 
```
I just ran the benchmarks on my GPU and the performance before the PR was indeed atrocious. After the changes we even start seeing a speedup on GPU compared to Numpy, which I assume will only get larger, the bigger the dataset. The Jax version also spends time compiling when first running a function, so this comparison might not be the fairest, depending on the use case.

src/ttsim/interface_dag_elements/fail_if.py
```diff
@@ -304,15 +304,12 @@ def input_data_is_invalid(input_data__flat: FlatData) -> None:
...
-            p_id_counts[i] += 1
-        else:
-            p_id_counts[i] = 1
+    # Convert to numpy array for performance (avoid slow JAX array iteration)
 
-    non_unique_p_ids = [i for i, count in p_id_counts.items() if count > 1]
+    # Use pandas duplicated for efficient duplicate detection
+    p_id_series = pd.Series(numpy.asarray(p_id))
+    duplicated_mask = p_id_series.duplicated(keep=False)
+    non_unique_p_ids = p_id_series[duplicated_mask].unique().tolist()
```
mj023, line 312: I think we can keep this in numpy, it should be similarly fast. I noticed that we are sorting the p_id array quite often, maybe we should at some point combine all these operations, so that we only sort once.

```suggestion
    p_id_sorted = numpy.sort(p_id)
    duplicates = numpy.where(numpy.diff(p_id_sorted, append=0) == 0, True, False)

    if numpy.sum(duplicates) >= 1:
        message = (
            "The following `p_id`s are not unique in the input data:\n\n"
            f"{p_id_sorted[duplicates]}\n\n"
        )
        raise ValueError(message)
```

mj023 (approved):

src/ttsim/interface_dag_elements/data_converters.py
JuergenWiemers: You are also right, the performance hit of removing `copy=False` is almost non-existent. I was more concerned about memory usage. The input data might be many GBs large, so we don't want to make unnecessary copies.

I just added (okay, Claude did) peak memory usage to `benchmark.py`: 
[benchmark.zip](https://github.com/user-attachments/files/21669057/benchmark.zip)

### WITHOUT `copy=False`
```
=======================================================================================================================================
MEMORY USAGE COMPARISON
=======================================================================================================================================
Households  Baseline    Data Size   NumPy Peak  JAX Peak    NumPy Inc   JAX Inc     NumPy Ratio JAX Ratio   Peak Ratio
(MB)        (MB)        (MB)        (MB)        (MB)        (MB)        (MB)        (Inc/Data)  (Inc/Data)  (JAX/NumPy)
---------------------------------------------------------------------------------------------------------------------------------------
32,767      133.1       42.5        186.7       269.3       11.1        91.2        0.3x        2.2x        1.44x
32,768      136.7       40.3        187.9       295.8       10.9        40.1        0.3x        1.0x        1.57x
65,536      137.3       82.0        240.7       371.7       21.3        45.1        0.3x        0.6x        1.54x
131,072     135.7       164.3       340.4       503.4       40.4        68.8        0.2x        0.4x        1.48x
262,144     135.6       327.1       543.8       735.6       81.1        113.2       0.2x        0.3x        1.35x
524,288     136.1       654.1       951.0       1179.7      160.8       209.2       0.2x        0.3x        1.24x
1,048,576   136.4       1308.1      1777.7      2047.5      333.3       397.9       0.3x        0.3x        1.15x
2,097,152   136.4       2616.1      3462.7      3654.3      710.2       666.9       0.3x        0.3x        1.06x
```

### WITH `copy=False`
```
=======================================================================================================================================
MEMORY USAGE COMPARISON
=======================================================================================================================================
Households  Baseline    Data Size   NumPy Peak  JAX Peak    NumPy Inc   JAX Inc     NumPy Ratio JAX Ratio   Peak Ratio
(MB)        (MB)        (MB)        (MB)        (MB)        (MB)        (MB)        (Inc/Data)  (Inc/Data)  (JAX/NumPy)
---------------------------------------------------------------------------------------------------------------------------------------
32,767      133.0       42.5        186.5       269.0       10.9        90.9        0.3x        2.2x        1.44x
32,768      136.5       40.4        187.3       294.4       10.3        37.1        0.3x        0.9x        1.57x
65,536      137.3       80.7        240.1       375.5       22.1        49.9        0.3x        0.6x        1.56x
131,072     135.4       164.6       344.2       507.0       44.2        70.4        0.3x        0.4x        1.47x
262,144     135.6       327.1       547.0       740.6       84.3        114.6       0.3x        0.4x        1.35x
524,288     136.1       654.1       966.7       1181.0      176.5       205.5       0.3x        0.3x        1.22x
1,048,576   136.7       1308.1      1790.4      2045.3      345.6       394.4       0.3x        0.3x        1.14x
2,097,152   136.8       2616.1      3457.9      3741.8      705.0       756.6       0.3x        0.3x        1.08x

Note: Peak Ratio > 1.0 means JAX uses more peak memory than NumPy
Note: 'Inc' = Memory increase during execution (peak - start)
Note: 'Ratio' = Memory increase relative to input data size
Note: Higher ratios indicate more memory overhead relative to input data
```

It seems `copy=False` doesn't really help very much here. 😅 So my worries were very much unfounded and it is basically costless to remove `copy=False` as you suggested.

src/ttsim/interface_dag_elements/fail_if.py
JuergenWiemers: ### With NumPy instead of Pandas:
```
================================================================================
Households  NumPy (s)   JAX (s)     NumPy Hash  JAX Hash    Speedup
--------------------------------------------------------------------------------
32,767      0.0860      0.8723      2584ff76    9a97a0ef    1/10.15x
32,768      0.0693      0.5594      a4ec6c3f    c0a7c7e9    1/8.08x
65,536      0.1166      0.6014      e18ee207    f873c7c8    1/5.16x
131,072     0.2135      0.7101      e949fd41    84608262    1/3.33x
262,144     0.4691      1.1576      a8fe5730    2fb136ba    1/2.47x
524,288     0.9420      2.0237      c7c95469    dd61a524    1/2.15x
1,048,576   1.9857      3.7792      1fbbd4ff    6ed4af0a    1/1.90x
2,097,152   4.8578      8.1802      81209a75    ff9f600f    1/1.68x
```

Exactly as fast as Pandas.

### With JAX instead of NumPy

I used this code:
```python
    # Check for non-unique p_ids
    p_id_sorted = xnp.sort(p_id)
    duplicates = xnp.where(xnp.diff(p_id_sorted, append=0) == 0, True, False)
    if xnp.sum(duplicates) >= 1:
        message = (
            "The following `p_id`s are not unique in the input data:\n\n"
            f"{p_id_sorted[duplicates]}\n\n"
        )
        raise ValueError(message)
```

```
================================================================================
PERFORMANCE COMPARISON
================================================================================
Households  NumPy (s)   JAX (s)     NumPy Hash  JAX Hash    Speedup
--------------------------------------------------------------------------------
32,767      0.0708      0.8176      2584ff76    9a97a0ef    1/11.54x
32,768      0.0659      0.6828      a4ec6c3f    c0a7c7e9    1/10.37x
65,536      0.1160      0.7421      e18ee207    f873c7c8    1/6.40x
131,072     0.2134      0.8519      e949fd41    84608262    1/3.99x
262,144     0.4294      1.1789      a8fe5730    2fb136ba    1/2.75x
524,288     0.8776      2.0219      c7c95469    dd61a524    1/2.30x
1,048,576   1.9762      3.8733      1fbbd4ff    6ed4af0a    1/1.96x
2,097,152   4.5925      7.9699      81209a75    ff9f600f    1/1.74x
```

Too close to call...

However, I just noticed that while all three approaches pass all tests in TTSIM, only the currently commited approach passes all GETTSIM tests. I get a lot of `ValueError: The following `p_id`s are not unique in the input data` errors in GETTSIM. It seems the GETTSIM tests cover some corner cases that the TTSIM-Tests do not?

hmgaudecker: Great analysis, thanks! 

Could you please
1. Add a test to TTSIM that makes the same error appear as in the GETTSIM test suite
2. Use the `xnp`-based implementation, adding an early return as below upfront?

```py
if p_id.shape[0] == 1:
    return
```

Reason is that before tackling that issue more generally (see comment above), I'd prefer to stick as close as possible to what we have rather than introducing a new strategy.

JuergenWiemers:
@hmgaudecker 
> 1. Add a test to TTSIM that makes the same error appear as in the GETTSIM test suite

I added a test that _would_ fail without the early return.:
```python
if p_id.shape[0] == 1:
    return
```

Without the early return, the test fails for the same reason that the GETTSIM tests failed: a one-row input dataset with `p_id==0` is falsely detected as a duplicate.

Is this the test you want?

hmgaudecker:
> Finally fixed the silly CI failures. Forgot to convert the Pandas `p_id` to NumPy/JAX. 🙈
> 
> But why does
> 
> ```python
> p_id_sorted = xnp.sort(p_id)
> ```
> 
> locally pass all tests (in a JAX-enabled Pixi environment) while it must be
> 
> ```python
> p_id_sorted = xnp.sort(xnp.asarray(p_id))
> ```
> 
> to pass GitHub's CI?

Are you running `pixi run tests-jax` ? Else you'll still use numpy as the backend, even if jax is installed.

hmgaudecker:
> Without the early return, the test fails for the same reason that the GETTSIM tests failed: a one-row input dataset with `p_id==0` is falsely detected as a duplicate.
> 
> Is this the test you thought about?

Precisely, many thanks! 

Basically, the situation we had should not happen: TTSIM test suite passes but GETTSIM test suite fails (essentially, GETTSIM tests should only be about policy contents, not about infrastructure). Given the recent split, this is to be expected though. However, we'll (almost) always want to create a TTSIM test then.

hmgaudecker (approved):
Wonderful, thanks a lot! 

I just combined the two tests into one parameterized test, will be clearer when it actually fails (e.g., you would see 5 failures instead of 2 with the loop-based implementation, so you get more information out). Also added the changelog.

PR #38, opened 2025-08-09
Plotting: Replace `nodes` selection type with `all_paths`

MImmesberger:

The `all_paths` selection type shows all paths that connect any primary nodes. There must be at least two primary nodes. Primary nodes that are not connected to any other primary node are omitted.

### DIFF
--- a/src/ttsim/plot/dag/tt.py
+++ b/src/ttsim/plot/dag/tt.py
@@ -297,7 +297,7 @@ def select_nodes_from_dag(
     selection_type: Literal["neighbors", "descendants", "ancestors", "all_paths"],
     selection_depth: int | None = None,
 ) -> nx.DiGraph:
-    """Select nodes based on the node selector."""
+    """Select nodes to plot."""
     if selection_type == "neighbors":
         order = selection_depth or 1
         selected_nodes = {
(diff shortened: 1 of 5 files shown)


src/ttsim/plot/dag/tt.py
```diff
@@ -65,14 +66,16 @@ def tt(
...
-        (for neighbors). If not provided, the entire DAG is plotted.
+        be root nodes (for descendants), end nodes (for ancestors), or middle nodes (for
+        neighbors). If not provided, the entire DAG is plotted.
     selection_type
         The type of the DAG to plot. Can be one of:
             - "neighbors": Plot the neighbors of the primary nodes.
             - "descendants": Plot the descendants of the primary nodes.
             - "ancestors": Plot the ancestors of the primary nodes.
-            - "nodes": Plot the primary nodes only.
+            - "all_paths": Plot all paths that connect primary nodes. Primary nodes are
```
hmgaudecker, line 76: I guess this is simply the behaviour of networkx? If easy, I think I'd prefer having all primary nodes in the plot, even if they are not connected to any of the others.

A specific test for this (either way that we end up choosing) would be good.

hmgaudecker (commented):
Looks great! I do have a question on that issue and I am still playing around with the GETTSIM notebook.

src/ttsim/plot/dag/tt.py
hmgaudecker: I'd prefer the orphaned nodes -- too easy to think you have done something wrong if nothing is shown. Agreed on the impurity in the naming, but that's the lesser evil IMO.

hmgaudecker (approved):
Great, thanks! Will just adjust the changelog to get rid of alpha Versions.

PR #42, opened 2025-08-17
Pre-sort user data by p_id for performance

JuergenWiemers:
Performance currently suffers if the user input data is not sorted by `p_id`.

### This PR:

- Adds `fail_if` functions that check `p_id` for existence and sanity (+ tests)
- Sorts the input data by the original `p_id` before creating the internal `p_id`. This happens in
  - `ttsim.interface_dag_elements.data_converters.df_with_mapped_columns_to_flat_data` and
  - `ttsim.interface_dag_elements.data_converters.df_with_nested_columns_to_flat_data` 
- Users get back the result DataFrame in the original sort order. This is achieved by storing an index column `__original_sort_indices__` along with the other data columns in the `FlatData` object returned by `df_with_mapped_columns_to_flat_data` and `df_with_nested_columns_to_flat_data`. Corresponding functions in `ttsim.interface_dag_elements.results.py` are adjusted to restore the original order.
- Some tests needed to be adjusted because they assumed that the `FlatData` objects might not be sorted by the original `p_id`.

### Performance effects

- In particular, the computation stage in NumPy strongly benefits from pre-sorting by original `p_id` in large datasets, but the JAX gains are also not too bad (see tables below).
- Performance for user data not sorted by the originial `p_id` is now essentially the same as if the user data had been passed already sorted.
- In the benchmarks, a hash mismatch in the computation stage is expected, since the sort order is different. Third stage hashes must match, however, because the original sort order is restored for the result DataFrames.
- @mj023: I tried to achieve further performance improvements by tinkering with the sort operations throughout the code, but the gains were minimal/non-existent. Sorting already sorted data seems to be basically a no-op.

#### Numpy backend:

```python
============================================================================================================================================
NUMPY BACKEND COMPARISON: Main Branch vs PR Branch - 3-STAGE BREAKDOWN
============================================================================================================================================
Households  Stage             Main (s)    PR (s)      Speedup     Description              Hash Match
--------------------------------------------------------------------------------------------------------------------------------------------
32,767      pre-processing    1.1365      1.1142      1.02x       Data preprocessing
            computation       0.8920      0.6038      1.48x       Core computation         ✗
            post-processing   1.0197      0.7827      1.30x       DataFrame formatting     ✓
            total time        3.0483      2.5007      1.22x       Complete execution
--------------------------------------------------------------------------------------------------------------------------------------------
32,768      pre-processing    1.1740      1.1032      1.06x       Data preprocessing
            computation       0.9136      0.6192      1.48x       Core computation         ✗
            post-processing   1.1063      0.8645      1.28x       DataFrame formatting     ✓
            total time        3.1938      2.5868      1.23x       Complete execution
--------------------------------------------------------------------------------------------------------------------------------------------
65,536      pre-processing    1.3000      1.0494      1.24x       Data preprocessing
            computation       1.7938      1.1627      1.54x       Core computation         ✗
            post-processing   0.8976      0.8177      1.10x       DataFrame formatting     ✓
            total time        3.9914      3.0298      1.32x       Complete execution
--------------------------------------------------------------------------------------------------------------------------------------------
131,072     pre-processing    1.5551      1.3531      1.15x       Data preprocessing
            computation       3.4427      2.3683      1.45x       Core computation         ✗
            post-processing   0.9128      0.9882      1/1.08x     DataFrame formatting     ✓
            total time        5.9106      4.7097      1.25x       Complete execution
--------------------------------------------------------------------------------------------------------------------------------------------
262,144     pre-processing    2.4494      1.9747      1.24x       Data preprocessing
            computation       7.4393      4.9451      1.50x       Core computation         ✗
            post-processing   0.8237      1.2887      1/1.56x     DataFrame formatting     ✓
            total time        10.7124     8.2085      1.31x       Complete execution
--------------------------------------------------------------------------------------------------------------------------------------------
524,288     pre-processing    7.9241      4.0708      1.95x       Data preprocessing
            computation       22.6316     11.3820     1.99x       Core computation         ✗
            post-processing   0.9392      1.8131      1/1.93x     DataFrame formatting     ✓
            total time        31.4949     17.2659     1.82x       Complete execution
--------------------------------------------------------------------------------------------------------------------------------------------
1,048,576   pre-processing    26.6471     8.5632      3.11x       Data preprocessing
            computation       70.6778     26.3814     2.68x       Core computation         ✗
            post-processing   1.6854      3.2799      1/1.95x     DataFrame formatting     ✓
            total time        99.0102     38.2245     2.59x       Complete execution
--------------------------------------------------------------------------------------------------------------------------------------------
```

#### JAX backend:

```python
============================================================================================================================================
JAX BACKEND COMPARISON: Main Branch vs PR Branch - 3-STAGE BREAKDOWN
============================================================================================================================================
Households  Stage             Main (s)    PR (s)      Speedup     Description              Hash Match
--------------------------------------------------------------------------------------------------------------------------------------------
32,767      pre-processing    4.9914      5.2813      1/1.06x     Data preprocessing
            computation       4.7661      4.4377      1.07x       Core computation         ✗
            post-processing   1.1202      1.2799      1/1.14x     DataFrame formatting     ✓
            total time        10.8777     10.9989     1/1.01x     Complete execution
--------------------------------------------------------------------------------------------------------------------------------------------
32,768      pre-processing    1.6930      1.8235      1/1.08x     Data preprocessing
            computation       4.5209      4.6488      1/1.03x     Core computation         ✗
            post-processing   0.9756      0.9961      1/1.02x     DataFrame formatting     ✓
            total time        7.1895      7.4684      1/1.04x     Complete execution
--------------------------------------------------------------------------------------------------------------------------------------------
65,536      pre-processing    1.9762      2.0121      1/1.02x     Data preprocessing
            computation       5.0280      4.7444      1.06x       Core computation         ✗
            post-processing   0.9757      1.0228      1/1.05x     DataFrame formatting     ✓
            total time        7.9799      7.7793      1.03x       Complete execution
--------------------------------------------------------------------------------------------------------------------------------------------
131,072     pre-processing    2.4043      2.3302      1.03x       Data preprocessing
            computation       6.2018      6.5097      1/1.05x     Core computation         ✗
            post-processing   0.9929      1.3359      1/1.35x     DataFrame formatting     ✓
            total time        9.5990      10.1758     1/1.06x     Complete execution
--------------------------------------------------------------------------------------------------------------------------------------------
262,144     pre-processing    3.4237      3.8534      1/1.13x     Data preprocessing
            computation       8.6577      7.3233      1.18x       Core computation         ✗
            post-processing   1.0553      1.1707      1/1.11x     DataFrame formatting     ✓
            total time        13.1367     12.3474     1.06x       Complete execution
--------------------------------------------------------------------------------------------------------------------------------------------
524,288     pre-processing    5.9624      5.7825      1.03x       Data preprocessing
            computation       14.8907     10.8211     1.38x       Core computation         ✗
            post-processing   1.0839      1.3854      1/1.28x     DataFrame formatting     ✓
            total time        21.9369     17.9890     1.22x       Complete execution
--------------------------------------------------------------------------------------------------------------------------------------------
1,048,576   pre-processing    11.9227     11.8409     1.01x       Data preprocessing
            computation       26.1555     17.8325     1.47x       Core computation         ✗
            post-processing   1.1681      2.4116      1/2.06x     DataFrame formatting     ✓
            total time        39.2462     32.0850     1.22x       Complete execution
--------------------------------------------------------------------------------------------------------------------------------------------
```

### DIFF
--- a/src/ttsim/interface_dag_elements/fail_if.py
+++ b/src/ttsim/interface_dag_elements/fail_if.py
@@ -579,6 +579,117 @@ def input_df_has_bool_or_numeric_column_names(
 
 
 @fail_function()
+def input_df_and_mapper_p_id_is_invalid(
+    input_data__df_and_mapper__df: pd.DataFrame,
+    input_data__df_and_mapper__mapper: NestedInputsMapper,
+) -> None:
+    """Fail if p_id column in DataFrame input is missing or invalid.
+
+    This validation runs before data conversion to ensure early failure.
+
+    Fails if:
+        - The mapper doesn't include a p_id mapping
+        - The p_id column is missing from the DataFrame
+        - The p_id column has non-integer values
+        - The p_id column has non-unique values
+    """
+    # Check if mapper includes p_id
+    mapper_flat = dt.flatten_to_qnames(input_data__df_and_mapper__mapper)
+    p_id_mapping = mapper_flat.get("p_id", None)
+
+    if p_id_mapping is None:
+        raise ValueError("The input mapper must include a mapping for 'p_id'.")
+
+    if not isinstance(p_id_mapping, str):
+        raise TypeError("The p_id mapping must be a string column name.")
+
+    # Check if p_id column exists in DataFrame
+    if p_id_mapping not in input_data__df_and_mapper__df.columns:
+        raise ValueError(
+            f"The p_id column '{p_id_mapping}' is missing from the DataFrame."
+        )
+
+    # Get p_id column data
+    p_id_series = input_data__df_and_mapper__df[p_id_mapping]
+
+    # Check data type
+    dtype_normalized = str(p_id_series.dtype).lower()
+    if "int" not in dtype_normalized:
+        msg = format_errors_and_warnings(
+            f"The p_id column must be of integer dtype. Got: {p_id_series.dtype}."
+        )
+        raise ValueError(msg)
+
+    # Check for uniqueness (skip for single-row data)
+    if len(p_id_series) > 1 and p_id_series.duplicated().any():
+        duplicated_ids = p_id_series[p_id_series.duplicated(keep=False)].unique()
+        raise ValueError(
+            f"p_id values must be unique. Duplicated values: {duplicated_ids.tolist()}"
+        )
+
+
+@fail_function()
+def input_df_with_nested_columns_p_id_is_invalid(
+    input_data__df_with_nested_columns: pd.DataFrame,
+) -> None:
+    """Fail if p_id column in nested DataFrame input is missing or invalid.
+
+    This validation runs before data conversion to ensure early failure.
+
+    Fails if:
+        - The ('p_id',) column is missing
+        - The p_id column has non-integer values
+        - The p_id column has non-unique values
+    """
+    # Find the p_id column efficiently (accounting for MultiIndex padding)
+    p_id_column = None
+    for col in input_data__df_with_nested_columns.columns:
+        # Handle both regular columns and MultiIndex tuples
+        if isinstance(col, tuple):
+            clean_col = tuple(el for el in col if not pd.isna(el))
+        else:
+            clean_col = (col,)
+
+        if clean_col == ("p_id",):
+            p_id_column = col
+            break
+
+    if p_id_column is None:
+        raise ValueError("Column with p_id data is missing from the DataFrame")
+
+    # Get p_id column data using the actual column name
+    try:
+        p_id_values = input_data__df_with_nested_columns[p_id_column]
+
+        # Handle MultiIndex DataFrame results
+        if hasattr(p_id_values, "values"):
+            p_id_array = numpy.asarray(p_id_values.values).flatten()  # Flatten to 1D
+        else:
+            p_id_array = p_id_values
+
+        # Create a Series for consistent interface
+        p_id_series = pd.Series(p_id_array)
+
+    except (KeyError, ValueError, AttributeError) as e:
+        raise ValueError(f"Could not access p_id column: {e}") from e
+
+    # Check data type
+    dtype_normalized = str(p_id_series.dtype).lower()
+    if "int" not in dtype_normalized:
+        msg = format_errors_and_warnings(
+            f"The p_id column must be of integer dtype. Got: {p_id_series.dtype}."
+        )
+        raise ValueError(msg)
+
+    # Check for uniqueness (skip for single-row data)
+    if len(p_id_series) > 1 and p_id_series.duplicated().any():
+        duplicated_ids = p_id_series[p_id_series.duplicated(keep=False)].unique()
+        raise ValueError(
+            f"p_id values must be unique. Duplicated values: {duplicated_ids.tolist()}"
+        )
+
+
+@fail_function()
 def input_df_mapper_columns_missing_in_df(
     input_data__df_and_mapper__df: pd.DataFrame,
     input_data__df_and_mapper__mapper: NestedInputsMapper,
--- a/src/ttsim/interface_dag_elements/results.py
+++ b/src/ttsim/interface_dag_elements/results.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 import dags.tree as dt
+import numpy
 import pandas as pd
 
 from ttsim.interface_dag_elements.data_converters import (
@@ -18,6 +19,41 @@ if TYPE_CHECKING:
         NestedStrings,
         QNameData,
     )
+
+
+def _restore_original_row_order(
+    df: pd.DataFrame,
+    input_data__flat: FlatData,
+) -> pd.DataFrame:
+    """Restore the original row order of a DataFrame using stored sort indices.
+
+    Args:
+        df: The DataFrame with rows in sorted order.
+        input_data__flat: The flat data containing the original sort indices.
+
+    Returns:
+        DataFrame with rows restored to their original order.
+    """
+    if ("__original_sort_indices__",) not in input_data__flat:
+        return df
+
+    sort_indices = input_data__flat[("__original_sort_indices__",)]
+
+    # Create inverse permutation: restore_order[orig_pos] = sorted_pos
+    restore_order = numpy.empty(len(sort_indices), dtype=int)
+    restore_order[sort_indices] = numpy.arange(len(sort_indices))
+
+    # Restore both the data rows AND the index values
+    index_name = df.index.name
+    original_index_values = df.index[restore_order]
+    df = df.iloc[restore_order]
+    df.index = original_index_values
+    df.index.name = index_name
+
+    if "__original_sort_indices__" in df.columns:
+        df = df.drop(columns=["__original_sort_indices__"])
+
+    return df
 
 
 @interface_function()
@@ -59,11 +95,14 @@ def df_with_mapper(
     -------
         A DataFrame.
     """
-    return nested_data_to_df_with_mapped_columns(
+    df = nested_data_to_df_with_mapped_columns(
         nested_data_to_convert=tree,
         nested_outputs_df_column_names=tt_targets__tree,
         data_with_p_id=input_data__flat,
     )
+
+    # Restore original row order if sort indices are available
+    return _restore_original_row_order(df, input_data__flat)
 
 
 @interface_function()
@@ -85,7 +124,10 @@ def df_with_nested_columns(
     -------
     A DataFrame with a hierarchical index in the column dimension.
     """
-    return nested_data_to_df_with_nested_columns(
+    df = nested_data_to_df_with_nested_columns(
         nested_data_to_convert=tree,
         index=pd.Index(input_data__flat[("p_id",)], name="p_id"),
     )
+
+    # Restore original row order if sort indices are available
+    return _restore_original_row_order(df, input_data__flat)
--- a/tests/interface_dag_elements/test_results.py
+++ b/tests/interface_dag_elements/test_results.py
@@ -1,0 +1,80 @@
+from __future__ import annotations
+
+import numpy
+import pandas as pd
+import pytest
+
+from ttsim.interface_dag_elements.results import _restore_original_row_order
+
+
+@pytest.mark.parametrize(
+    (
+        "df",
+        "input_data__flat",
+        "expected_row_order",
+        "expected_index",
+        "should_be_unchanged",
+    ),
+    [
+        # No sort indices - should return DataFrame unchanged
+        (
+            pd.DataFrame(
+                {"col1": [10, 20, 30], "col2": [100, 200, 300]},
+                index=pd.Index([3, 1, 2], name="p_id"),
+            ),
+            {},  # No sort indices
+            [10, 20, 30],  # Should remain unchanged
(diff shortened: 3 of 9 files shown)


src/ttsim/interface_dag_elements/processed_data.py
```diff
@@ -34,30 +34,38 @@ def processed_data(input_data__flat: FlatData, xnp: ModuleType) -> QNameData:
...
+                sorted_internal_p_ids[insert_positions],
+                sorted_data_array,
             )
             processed_input_data[qname] = variable_with_new_ids
         else:
-            processed_input_data[qname] = xnp.asarray(data)
+            processed_input_data[qname] = sorted_data_array
+
+    # Store original sort indices for row order restoration in results
+    processed_input_data["__original_sort_indices__"] = sort_indices
```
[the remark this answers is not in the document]
hmgaudecker: Actually, taking a closer look I think it should only be something like:

```py
def sort_indices(flat: FlatData, xnp: ModuleType) -> IntColumn:
    return xnp.argsort(xnp.asarray(input_data__flat[("p_id",)]))
```

And then use this in a minimally-invasive fashion as an input to `processed_data` and to `results__tree`. 

Please get rid of the `raw_results__combined` function -- now that the raw results have different sort order in `columns` and in `from_input_data`, this function does not make sense any more. We should do all the reverting-back-to-original-style in `results`. So `results_tree` should take the three inputs of `raw_results__combined` instead of that function.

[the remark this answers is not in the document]
hmgaudecker: Yeah, we'll need 1.1 for sure. Though outside of benchmarks, I am pretty sure nobody was requesting raw results.

Note, however, that returning "combined" raw results with different sort orders would have been much worse despite the interface looking the same!

src/ttsim/interface_dag_elements/results.py
```diff
@@ -14,29 +15,94 @@
...
+    input_data__sort_indices: IntColumn,
+) -> NestedData:
+    """The combined results as a tree with original row order restored.
 
-    The transformed id's are converted back to their original values.
+    The transformed id's are converted back to their original values, and the
+    row order is restored to match the original input data order.
 
     """
+    # Combine the three result sources
```
hmgaudecker, line 82: I am pretty sure this will benefit from a helicopter view. 

In particular, we know all the structure that is being dissected in `_restore_original_row_order_in_nested_data` from the different types of inputs. So I am pretty sure we should be able to do without the function call and just do the re-sorting on the fly. 

Would something like the following (untested) code work?

```py
    out = {
        **raw_results__params,
        **raw_results__from_input_data,
    }

    # Create inverse permutation: if sort_indices[i] = j, then restore_order[j] = i
    restore_order = numpy.empty(len(input_data__sort_indices), dtype=int)
    restore_order[input_data__sort_indices] = numpy.arange(
        len(input_data__sort_indices)
    )
    for k, v in raw_results__columns.items():
        path = dt.tree_path_from_qname(k)
        if path in input_data__flat and (
            path[-1].endswith("_id") or path[-1].startswith("p_id_")
        ):
            out[k] = input_data__flat[path]
        else:
            out[k] = restore_order[v]

    return dt.unflatten_from_qnames(out)
```

Staring at the code, I even think the case distinction is superfluous. I.e., if something is in the `input_data`, it must not be among the column results. But I might be taking a shortcut in my brain, @MImmesberger can tell you a thing or two about such paths... But we should definitely try this:

```py

    # Create inverse permutation: if sort_indices[i] = j, then restore_order[j] = i
    restore_order = numpy.empty(len(input_data__sort_indices), dtype=int)
    restore_order[input_data__sort_indices] = numpy.arange(
        len(input_data__sort_indices)
    )

    return dt.unflatten_from_qnames(
        {
            **raw_results__params,
            **raw_results__from_input_data,
            **{k: restore_order[v] for  k, v in raw_results__columns.items()},
        }
    )
```

src/ttsim/interface_dag_elements/results.py
```diff
@@ -49,20 +115,32 @@ def df_with_mapper(
...
     Args:
         tree:
-            The results of a TTSIM run.
-        input_data__tree:
-            The data tree of the TTSIM run.
-        nested_outputs_df_column_names:
+            The results of a TTSIM run with original row order already restored.
+        input_data__flat:
+            The input data containing original p_ids.
+        tt_targets__tree:
```
hmgaudecker, line 121: With the exception of the docstring updates (thanks!!!), this function should stay the same as it was. Or am I missing something?

src/ttsim/interface_dag_elements/results.py
```diff
@@ -75,17 +153,26 @@ def df_with_nested_columns(
 
     Args:
         tree:
-            The results of a TTSIM run.
-        input_data__tree:
-            The data tree of the TTSIM run.
-        nested_outputs_df_column_names:
-            A tree that maps paths (sequence of keys) to data columns names.
+            The results of a TTSIM run with original row order already restored.
```
hmgaudecker, line 156: Same.

hmgaudecker (commented):
Great! Almost there!

src/ttsim/interface_dag_elements/results.py
JuergenWiemers: I think this is just a leftover from my earlier implementation in which I got weird test failures here. Now everything works fine with the untouched original version.

JuergenWiemers:
- GETTSIM tests still good
- Performance still good
- Will update my benchmark scripts in "code for picking" once this is merged

src/ttsim/interface_dag_elements/results.py
```diff
@@ -14,29 +15,36 @@
...
-            path[-1].endswith("_id") or path[-1].startswith("p_id_")
-        ):
-            out[k] = input_data__flat[path]
-        else:
-            out[k] = raw_results__combined[k]
-    return dt.unflatten_from_qnames(out)
+    return dt.unflatten_from_qnames(
+        {
+            **raw_results__params,
+            **{k: reorder_arrays(v) for k, v in raw_results__from_input_data.items()},
```
hmgaudecker, line 44: Wait, why do we need to reorder the input data? They should have never changed the order in the first place, I thought? 

Can you double-check by adding some column from the input data to the tt_targets of some end-to-end tests?

tests/interface_dag_elements/test_results.py
```diff
@@ -0,0 +1,91 @@
...
+        "raw_results_columns",
+        "raw_results_params",
+        "raw_results_from_input_data",
+        "input_data__sort_indices",
+        "expected_result",
+    ),
+    [
+        (
+            # Raw results columns with sorted data
+            {"result_col": numpy.array([200, 300, 100])},
```
hmgaudecker, line 20: Can you add test cases 
```suggestion
            {"result_col": numpy.array([200, 300, 100]), "result_scalar": 1.0},
```

Also add the same thing to `raw_results_from_input_data` and some dict or so to `raw_results_from_params`, so we check all cases that we made provisions for.

hmgaudecker (commented):
Last mile!

src/ttsim/interface_dag_elements/results.py
hmgaudecker: Sorry, strike that! 

You are completely right, ofc, confused myself that `raw_results__from_input_data` uses `processed_data` as an input. 

Terrible naming though.

@MImmesberger, should we rename to `from_processed_data` ? It is not really user-facing, so keeping the internals straight seems the greater good relative to using the intuitive name to the outside.

src/ttsim/interface_dag_elements/results.py
```diff
@@ -14,29 +15,36 @@
...
+    input_data__sort_indices: IntColumn,
+) -> NestedData:
+    """The combined results as a tree with original row order restored."""
+    restore_order = numpy.empty(len(input_data__sort_indices), dtype=int)
+    restore_order[input_data__sort_indices] = numpy.arange(
+        len(input_data__sort_indices)
+    )
 
-    The transformed id's are converted back to their original values.
+    def reorder_arrays(v: object) -> object:
```
hmgaudecker, line 38:
```suggestion
    def reorder_arrays(v: Any) -> Any:
```
Just because we are using this one in other places of the code.

src/ttsim/interface_dag_elements/results.py
```diff
@@ -14,29 +15,36 @@
...
         NestedData,
         NestedStrings,
         QNameData,
     )
 
 
 @interface_function()
-def tree(raw_results__combined: QNameData, input_data__flat: FlatData) -> NestedData:
-    """The combined results as a tree.
+def tree(
```
hmgaudecker, line 26: Looking at the below, I think we want to add a function:

`fail_if__p_id_x_among_targets`

That is, unless I am missing something we currently do not support the case where we set up some `p_id_whatever` endogenously. It will continue pointing to the internal `p_id` by the below machinery. No change in behaviour due to this PR (it just made this clear!), but it is a problem.

I do not think we should implement it (find it hard to think of an example), but we should not return wrong results silently.

src/ttsim/interface_dag_elements/results.py
JuergenWiemers: Ah, got it, just needed to add `  # noqa: ANN401`

hmgaudecker: Just append `# noqa: ANN401` to the end of the line. Should be plenty of examples in the codebase.

src/ttsim/interface_dag_elements/results.py
JuergenWiemers: I over-eagerly implemented this before I read "I do not think we should implement it". I'll just revert it.

src/ttsim/interface_dag_elements/fail_if.py
```diff
@@ -992,3 +992,38 @@ def param_function_depends_on_column_objects(
...
+    ]
+
+    if p_id_targets:
+        formatted = format_list_linewise(p_id_targets)
+        msg = (
+            "The following p_id_* columns were requested as targets, but these "
+            "contain internal ID mappings that are not meaningful for users:\n\n"
+            f"{formatted}\n\n"
+            "Please remove these from your targets specification. If you need person "
+            "identifiers in your results, use the original ID columns from your input data."
```
hmgaudecker:
```suggestion
            "identifiers in your results, merge the original ID columns from your input data to the result."
            "\nIn case you need this behaviour, please add your request to https://github.com/ttsim-dev/ttsim/issues/XXX"
```
Definitely keep this function, see my other comment..

JuergenWiemers:
By the way, I'm thinking about striking "for performance" from the PR title and the entry in `CHANGES.md`. When I started working on this, I really thought about it as a pure performance optimization (that's all I ever do 😉), but now it has evolved into a compromise between user-friendliness and performance. If we really only cared about performance, the best way to go would be an additional check for sorting by the original `p_id` in `fail_if.input_data_is_invalid` and tell the user "Just sort your data already! (If you are a Stata user, add the line `isid <YOUR_PID_NAME>, sort` just before saving the dataset.) Thank you." Then all the shenanigans we do right now (sorting, preserving the reordering information throughout the lifetime of the simulation, reodering the tree columns for output) wouldn't be necessary. But, of course, user-friendliness is important. 😃

PR #52, opened 2025-09-04
Fix input data and policy environment `fail_if` checks.

MImmesberger:

This PR does two things:
- `fail_if__input_data_tree_is_invalid` was too strict because it allowed for array leafs only. It now also allows for float/int/bool. Also fixed some corresponding bugs down the line.
- `fail_if__environment_is_invalid` was too strict because it did not allow for setting `evaluation_year, evaluation_date, evaluation_month` as scalars, even though those are `PolicyInput`s in the policy environment.

### DIFF
--- a/src/ttsim/interface_dag_elements/fail_if.py
+++ b/src/ttsim/interface_dag_elements/fail_if.py
@@ -35,6 +35,7 @@ from ttsim.tt.param_objects import (
     PLACEHOLDER_FIELD,
     PLACEHOLDER_VALUE,
     ParamObject,
+    ScalarParam,
 )
 
 if TYPE_CHECKING:
@@ -258,7 +259,7 @@ def input_data_tree_is_invalid(
     input_data__tree: NestedData, backend: Literal["numpy", "jax"], xnp: ModuleType
 ) -> None:
     """Validate the basic structure of the input data tree."""
-    valid_leaf_types = (pd.Series, numpy.ndarray, xnp.ndarray)
+    valid_leaf_types = (pd.Series, numpy.ndarray, xnp.ndarray, int, float, bool)
     if backend == "numpy" and jax is not None:
         valid_leaf_types = (*valid_leaf_types, jax.numpy.ndarray)
     assert_valid_ttsim_pytree(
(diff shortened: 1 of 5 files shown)


CHANGES.md
```diff
@@ -4,6 +4,11 @@ This is a record of all past `ttsim` releases and what went into them in reverse
 chronological order. We follow [semantic versioning](https://semver.org/) and all
 releases are available on [Anaconda.org](https://anaconda.org/conda-forge/ttsim).
 
+## Unreleased
+
+- {gh}`52` Fix input data and policy environment `fail_if` checks
+  ({ghuser}`MImmesberger`)
+
 ## v1.1.1 — unpublished
```
hmgaudecker, line 12:
```suggestion
```
Doubled up.

src/ttsim/interface_dag_elements/fail_if.py
```diff
@@ -303,11 +304,13 @@ def input_data_is_invalid(input_data__flat: FlatData, xnp: ModuleType) -> None:
         )
         raise ValueError(message)
 
+    # Check that all arrays have the same length as the p_id array.
     len_p_id_array = len(input_data__flat[("p_id",)])
```
hmgaudecker, line 308:
```suggestion
    if not hasattr(input_data__flat[("p_id",)], "__len__"):
        raise ValueError("`p_id` must be an array or series.")
    len_p_id_array = len(input_data__flat[("p_id",)])
```
Else this will give "object of type int has no len" or so.

src/ttsim/interface_dag_elements/fail_if.py
```diff
@@ -322,16 +325,46 @@ def environment_is_invalid(
     policy_environment: PolicyEnvironment,
```
hmgaudecker, line 325: Please rename check to `policy_environmen_is_invalid`, I was quite confused when looking at the PR.

Looks like I was not the only one -- please revert the changes to this check, neither the backend nor the evaluation date belongs into the policy environment (see the interface DAG).

CHANGES.md
```diff
@@ -4,6 +4,11 @@ This is a record of all past `ttsim` releases and what went into them in reverse
 chronological order. We follow [semantic versioning](https://semver.org/) and all
 releases are available on [Anaconda.org](https://anaconda.org/conda-forge/ttsim).
 
+## Unreleased
```
hmgaudecker, line 7:
```suggestion
## v1.1.1 — unpublished
```
All this still counts as a bug, I'd say.

hmgaudecker (commented):
I think the title should become "Fix `fail_if` checks: input data tree was too greedy and environment had an ambiguous name."

hmgaudecker (approved):
Very nice! Could you just add a test passing a `p_id` scalar? Just in the spirit of keeping coverage high...

PR #53, opened 2025-09-04
Partialling in scalars from input data doesn't work for vectorized functions

MImmesberger:

In TTSIM we silently assume that "columns" are arrays, not scalars. This leads to errors when passing scalar data. For example:

- Many `group_creation_function`s use array methods (e.g. when sorting). For example, when using METTSIM and set `age` to a fixed scalar, `sorted_age = age[sorting]` in `fam_id` will fail.
- In `grouped_sum` we have `if column.dtype in ["bool"]` that fails if `column` is not an array.
- Probably all other places were we turned off vectorization are problematic too

To resolve these, we could create arrays from these scalars that match the length of `p_id`. However, it seems quite cumbersome to do this manually for each vectorized function. I think it would be much nicer to take care of this automatically. 

I think it would be quite straightforward to work with the original type hints (before vectorization) to find out which columns should be arrays (or we could write that down somewhere in the decorator). However, when partialling in those scalars/arrays, we don't know the length of `p_id`. 

Any ideas @hmgaudecker?

### Todo

- [ ] Pick an appropriate title.
- [ ] Put `Closes #XXXX` in the first PR comment to auto-close the relevant issue once
      the PR is accepted. This is not applicable if there is no corresponding issue.
- [ ] Document PR in CHANGES.md.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -37,8 +37,8 @@ jobs:
           - os: windows-latest
             environment: py312
     steps:
-      - uses: actions/checkout@v4
-      - uses: prefix-dev/setup-pixi@v0.8.14
+      - uses: actions/checkout@v5
+      - uses: prefix-dev/setup-pixi@v0.9.0
         with:
           pixi-version: v0.50.1
           cache: true
--- a/.github/workflows/publish-to-pypi.yaml
+++ b/.github/workflows/publish-to-pypi.yaml
@@ -8,7 +8,7 @@ jobs:
     permissions:
       id-token: write
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - name: Set up Python 3.12
         uses: actions/setup-python@v5
         with:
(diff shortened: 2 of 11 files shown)


MImmesberger:
Yes, I agree. Another solution is just to provide (some of) the aggregations and group IDs yourself as scalars. Only works if they are not supposed to depend on an endogenous column, though.

> A check that takes the processed data and one of the specialized environments before vectorization should do, right?

Meaning: Loop thorugh column functions and raise error if it has `vectorization_strategy="not_required"` and one of its arguments is a scalar in processed data? That should work.

hmgaudecker:
> Meaning: Loop thorugh column functions and raise error if it has `vectorization_strategy="not_required"` and one of its arguments is a scalar in processed data? That should work.

1. Make sure `tt_function_set_annotations` is set
2. Create DAG from `("specialized_environment", "with_partialled_params_and_scalars")` 
3. Loop through root nodes that are `(Bool|Int|Float)Columns` and check that they are not scalars in the processed data

We might be a little too greedy than absolutely necessary, but this is really a corner case and just a(n important) convenience for users.

hmgaudecker (approved):
Looks great, thanks!

PR #55, opened 2025-09-14
Fix _q to _m time conversion.

MImmesberger:

The _q to _m automatic time conversion (and the corresponding test) was incorrect. This PR fixes that.


hmgaudecker (approved):
Ooops. Great catch, thanks!

PR #63, opened 2025-12-22
Move from mypy to ty

hmgaudecker:
Lots of improvements, highlights:

- The typing of `OrigParamSpec` was off. What we had written was "one of two exclusive kinds of dictionaries" where we wanted to write "One dictionaries where key/value pairs can take one of two forms". Fixed via a Protocol.
- Use `TypeAlias` for `OrderedQNames` instead of `TypeVar` (which created a different instance per function / module).
- Define `RawParamValue` since we were using `RawParam` as a type annotation in many places where `RawParam.value` was passed.
- No global overrides! *(Except for `invalid-return-type`, which is required for `policy_inputs`, where we annotate with the expected type, but return `None`)*

ToDo:
- [x] Behavior of `_fail_if_requested_nodes_cannot_be_found` was incorrect and tests only passed because we happened to add the item that should fail at the end, else earlier failures would get overwritten by the loop. See demonstration in this [commit](https://github.com/ttsim-dev/ttsim/pull/63/commits/9bf7826885d919cf8f83afc5518c57fbfa204355). Check relevant changes: `git diff 12be01be395dc4db6c6f119ff3c11d0a73dd0416~2`.

  When fixing the loop and not commenting out the second clause below, we saw failures.

  Issue is that when we request only upstream stuff, input-dependent interface functions like `(input_data, flat)` are never being created. It is not fully obvious to me why we need that check at all. 

```py
     if missing_main_targets:  # or missing_main_targets_from_include_conditions:
```
- [x] When done with the above, remove `"tests/test_main.py" = ["ERA001"]` from `pyproject.toml`

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -40,7 +40,7 @@ jobs:
       - uses: actions/checkout@v6
       - uses: prefix-dev/setup-pixi@v0.9.3
         with:
-          pixi-version: v0.61.0
+          pixi-version: v0.62.2
           cache: true
           cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
           frozen: true


src/ttsim/plot/dag/interface.py
```diff
@@ -135,6 +135,8 @@ def interface(
                 else interface_object
             )
             description = inspect.getdoc(f)
+        else:
+            description = ""
```
MImmesberger:
```suggestion
            description = "No description available."
```

And remove the default below.

MImmesberger (approved):
Cool, I think this makes the type hints much more reliable, which will be incredibly useful!

.pre-commit-config.yaml
```diff
@@ -72,22 +72,14 @@ repos:
...
-          - --allow-redefinition-new
-          - --local-partial-types
-        additional_dependencies:
-          - types-PyYAML
-          - types-pytz
-          - numpy >= 2
-          - jaxtyping
-          - dags >= 0.4.1
-          - optree >= 0.16
+  # - repo: local
```
MImmesberger: Flagging because commented out

src/ttsim/interface_dag_elements/templates.py
```diff
@@ -34,9 +34,9 @@ def input_data_dtypes__tree(
     """A template of the required input data and their expected types."""
     base_dtype_tree = dt.create_tree_with_input_types(
         functions=dt.unflatten_from_qnames(
-            specialized_environment_for_plotting_and_templates__with_partialled_params_and_scalars,
```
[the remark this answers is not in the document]
hmgaudecker: Unfortunately, the two are not nested indeed. We could work more with custom things, but for now I'd just wait for the typing landscape to evolve a bit.

src/ttsim/plot/dag/interface.py
hmgaudecker: Thx!

PR #67, opened 2026-01-17
Better selection mechanism for `node.colormap` in graphs.

### LINKED ISSUE #45: ENH: Let `plot.dag.tt` colormap pick up variations of policy objects

After #44, we can specify colors at the level of namespaces and individual objects. What is missing is a middle ground. This is particularly relevant for the (heterogeneous) top-level namespace, where one may want to group all `p_id`s in one group, wealth-related variables in another, etc. Currently, one would have to specify these one by one, which is particularly annoying when automatic time conversions / aggregations come into play.

Two possible solutions:

1. Extend to all time conversions / aggregations where relevant
2. Use a regex-based solution, e.g. `(r"wealth.+",)` would catch all variations of wealth in the top-level namespace

hmgaudecker:
Fix #45 using standard glob patterns:

| Pattern element | Meaning |
|-----------------|---------|
| `"exact_name"` | Matches this exact path segment |
| `"prefix*"` | Matches segments starting with `prefix` |
| `"*suffix"` | Matches segments ending with `suffix` |
| `"*infix*"` | Matches segments containing `infix` |
| `"?"` | Matches exactly one character |
| `"**"` | Matches any number of path segments (including zero) |
| `"top-level"` | Special: matches any single-segment path (top-level nodes) |

- In action [here](https://gettsim--1128.org.readthedocs.build/en/1128/tutorials/visualizing_the_system.html#Glob-patterns-for-flexible-matching).
- Twin [GETTSIM #1128](https://github.com/ttsim-dev/gettsim/pull/1128)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,7 +9,7 @@ releases are available on [Anaconda.org](https://anaconda.org/conda-forge/ttsim)
 - {gh}`67` Add glob-style pattern matching for `node_colormap` in DAG plotting.
   ({ghuser}`hmgaudecker`)
 
-## v1.1.0 — 2026-01-15
+## v1.1.1 — 2026-01-12
 
 - {gh}`65` Add Python 3.14 support. ({ghuser}`hmgaudecker`)
 - {gh}`63` Use ty as the type checker for ttsim. Many improvements to type safety.


MImmesberger (approved):
Looks good to me, don't forget the CHANGES.md

One minor thing: The color map is defined over paths, while the primary nodes selector of the plotting function accepts qualified names; in fact, we mostly use qnames for plotting in the documentation. I think we should allow for qnames in the color map too.

MImmesberger (approved):

PR #71, opened 2026-01-21
Fix pygraphviz plugin registration on macOS

MImmesberger:
Runs `dot -c` before pytest on macOS to register graphviz plugins. This was previously done manually after each `pixi install`; now it's automated in the test tasks.

Also removed the `dot -c` config via GitHub Actions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,20 @@
+# Build artifacts
+*.egg-info
+_version.py
+bld
+build
+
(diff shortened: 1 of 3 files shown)


MImmesberger:
@hmgaudecker Just saw we were both working on the pyproject toml at the same time. I changed the base of this PR from main to `boilerplate`.

PR #83, opened 2026-02-12
Allow non-string annotations in vectorization.

hmgaudecker:

Internally, we need string annotations (via `from __future__ import annotations`) and `if TYPE_CHECKING` blocks in order to not require jax as a runtime dependency:

- Bool, Float, Int (dtype validators) can be imported without JAX
- Array cannot - it triggers import jax via a `__getattr__` hook in jaxtyping's `__init__.py`

Since the column types are defined as `IntColumn = Int[Array, " n_obs"]`, they need `Array`, which needs JAX. So we can't move them out of `TYPE_CHECKING` without making JAX a hard runtime dependency.

The registry has to keep returning strings. And `from __future__ import annotations` stays necessary in files using these column type annotations.

However, there is no reason to require this from user code using our vectorization machinery. Changing two lines helps.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -12,6 +12,21 @@ on:
     branches:
       - '*'
 jobs:
+  run-ty:
+    name: Run ty
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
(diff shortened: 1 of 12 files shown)


MImmesberger (approved):
Nice!

PR #92, opened 2026-03-21
Fix loop vectorization for functions with param helpers

MImmesberger:
## Summary

Vectorisation strategy `"loop"` failed if functions had piecewise polynomial param inputs. This PR fixes this.

Related to #91 — these fixes enable downstream projects (e.g. GETTSIM) to switch policy functions to the `"loop"` strategy for any (not already natively) vectorised function.

- `look_up()` and `piecewise_polynomial()` now return numpy scalars for scalar input, instead of always returning arrays. This makes them compatible with `numpy.vectorize`, which calls functions per-element and expects scalar returns.
- The `"loop"` path in `vectorize_function` now wraps `numpy.vectorize` in a regular function closure instead of applying `functools.wraps` directly to the `numpy.vectorize` object. This fixes `AnnotationMismatchError` in dags, which could not resolve type annotations on `numpy.vectorize` objects.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,11 @@ This is a record of all past `ttsim` releases and what went into them in reverse
 chronological order. We follow [semantic versioning](https://semver.org/) and all
 releases are available on [Anaconda.org](https://anaconda.org/conda-forge/ttsim).
 
+## Unreleased
+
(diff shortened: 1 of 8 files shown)


hmgaudecker (approved):
Very nice, thanks!

PR #101, opened 2026-05-22
beartype + jaxtyping runtime type-checking rollout

### LINKED ISSUE #39: Error from automatic bool aggregation type hints could be improved when the target is a bool
### Bug description

Using GETTSIM, I accidentally defined `wohnort_ost` instead of `wohnort_ost_hh` (which would be the correct input) in my input data. 

Currently, this throws the following error
```shell
File ~/GitHub/gettsim/.pixi/envs/default/lib/python3.13/site-packages/ttsim/interface_dag_elements/tt_function.py:43, in tt_function(specialized_environment__tt_dag, specialized_environment__with_partialled_params_and_scalars, labels__column_targets, tt_function_set_annotations, backend)
     33 @interface_function(in_top_level_namespace=True)
     34 def tt_function(
     35     specialized_environment__tt_dag: nx.DiGraph,
   (...)     39     backend: Literal["numpy", "jax"],
     40 ) -> Callable[[QNameData], QNameData]:
     41     """Returns a function that takes a dictionary of arrays and unpacks them as keyword
     42     argugments."""
---> 43     ttf_with_keyword_args = concatenate_functions(
     44         dag=specialized_environment__tt_dag,
     45         functions=specialized_environment__with_partialled_params_and_scalars,
     46         targets=list(labels__column_targets),
     47         return_type="dict",
     48         aggregator=None,
     49         enforce_signature=True,
     50         set_annotations=tt_function_set_annotations,
     51     )
     53     if backend == "jax":
     54         import jax  # noqa: PLC0415

File ~/GitHub/gettsim/.pixi/envs/default/lib/python3.13/site-packages/dags/dag.py:172, in concatenate_functions(functions, targets, dag, return_type, aggregator, enforce_signature, set_annotations, lexsort_key)
    169     dag = create_dag(functions=functions, targets=targets)
    171 # Build combined function.
--> 172 return _create_combined_function_from_dag(
    173     dag=dag,
    174     functions=functions,
    175     targets=targets,
    176     return_type=return_type,
    177     aggregator=aggregator,
    178     enforce_signature=enforce_signature,
    179     set_annotations=set_annotations,
    180     lexsort_key=lexsort_key,
    181 )

File ~/GitHub/gettsim/.pixi/envs/default/lib/python3.13/site-packages/dags/dag.py:295, in _create_combined_function_from_dag(dag, functions, targets, return_type, aggregator, enforce_signature, set_annotations, lexsort_key)
    288 _exec_info = create_execution_info(
    289     _functions, dag, verify_annotations=set_annotations, lexsort_key=lexsort_key
    290 )
    292 # Create the concatenated function that returns all requested targets as a tuple.
    293 # If set_annotations is True, the return annotation is a tuple of strings,
    294 # corresponding to the return types of the targets.
--> 295 _concatenated = _create_concatenated_function(
    296     _exec_info,
    297     _arglist,
    298     _targets,
    299     enforce_signature,
    300     set_annotations,
    301 )
    303 # Update the actual return type, as well as the return annotation of the
    304 # concatenated function.
    305 out: Callable[..., Any]

File ~/GitHub/gettsim/.pixi/envs/default/lib/python3.13/site-packages/dags/dag.py:593, in _create_concatenated_function(execution_info, arglist, targets, enforce_signature, set_annotations)
    590 return_annotation: type[inspect._empty] | tuple[str, ...]
    592 if set_annotations:
--> 593     args, return_annotation = get_annotations_from_execution_info(
    594         execution_info,
    595         arglist=arglist,
    596         targets=targets,
    597     )
    598 else:
    599     args = arglist

File ~/GitHub/gettsim/.pixi/envs/default/lib/python3.13/site-packages/dags/dag.py:685, in get_annotations_from_execution_info(execution_info, arglist, targets)
    682     types.update(info.argument_annotations)
    684 if errors:
--> 685     raise AnnotationMismatchError(
    686         "The following type annotations are inconsistent:\n" + "\n".join(errors)
    687     )
    689 args_annotations = {arg: types[arg] for arg in arglist}
    690 return_annotation = tuple(types[target] for target in targets)

AnnotationMismatchError: The following type annotations are inconsistent:
function sozialversicherung__rente__beitrag__beitragsbemessungsgrenze_m has the argument type annotation 'wohnort_ost_hh: BoolColumn', but function wohnort_ost_hh has return type: FloatColumn | IntColumn.
```

We should either:
- disallow automatic bool aggregations explicitly (and throw a better error)
- be more lenient with the type hint (i.e. add BoolColumn)

hmgaudecker:
## Summary

Rolls out **beartype** runtime type-checking across ttsim, with **jaxtyping** shape-tagged array types as the single source of truth for annotations.

- Package-wide claw (`beartype_package("ttsim")`), env-var-gated on `TTSIM_BEARTYPE_CLAW` for the rollout; CI + pixi tests env set it to `1`.
- `ttsim.exceptions` hierarchy rooted at `TTSIMError`; per-component `BeartypeConf` instances in `ttsim._beartype_conf` map violations to typed errors at user boundaries.
- Wide/narrow type split (`User*` boundary aliases vs. canonical column aliases); backend-agnostic `Array | np.ndarray` unions; runtime-resolvable typing aliases.
- jaxtyping ellipsis-sentinel pickle patch; build-time DAG type-resolution module.
- All claw-surfaced failures peeled — including the interface-DAG `InterfaceFunction` bound-method corruption (tests now load interface objects claw-free, the way the DAG does).

## Test plan

- [x] `TTSIM_BEARTYPE_CLAW=1` full suite: **986 passed, 0 failed**
- [x] claw-off suite green
- [x] `ty` clean, `prek` clean

## Before merge

- Confirm the `dags` dependency pin is release-appropriate (currently tracks `dags` main).


hmgaudecker (commented):
Autoreview

src/ttsim/interface_dag_elements/tt_targets.py
```diff
@@ -38,6 +41,6 @@ def tree(policy_environment: PolicyEnvironment) -> NestedTargetDict | NestedStri
 
 
 @interface_function()
-def qname(tree: NestedTargetDict | NestedStrings) -> OrderedQNames:
+def qname(tree: NestedTargetDict | NestedStrings) -> FlatTTTargets:
```
MImmesberger: We should rename `FlatTTTargets`. Throughout the codebase `flat` always means a dict with tree paths as keys IIRC

Another violation is `FlatColumnObjects`.

src/ttsim/plot/dag/interface.py
```diff
@@ -6,6 +6,10 @@
 import dags
 import dags.tree as dt
 
+# Hoisted out of `TYPE_CHECKING` so the beartype claw can resolve the
```
MImmesberger: I think this comment should go, our default should be to move as much stuff as possible outside of the `TYPE_CHECKING` block (IIRC we only need it for JAX).

src/ttsim/plot/dag/tt.py
```diff
@@ -19,18 +26,14 @@
...
-    import plotly.graph_objects as go
-
-    from ttsim.main_args import InputData, Labels
-    from ttsim.typing import (
-        DashedISOString,
-        PolicyEnvironment,
-        QNameData,
-        SpecEnvWithoutTreeLogicAndWithDerivedFunctions,
-    )
+# Hoisted out of `TYPE_CHECKING` so the beartype claw can resolve these
```
MImmesberger: See above.

src/ttsim/tt/aggregation_jax.py
```diff
@@ -2,6 +2,10 @@
 
 from typing import TYPE_CHECKING
 
+# Hoisted out of `TYPE_CHECKING` so the beartype claw can resolve these
```
MImmesberger: See above

src/ttsim/interface_dag_elements/fail_if.py
```diff
@@ -231,7 +238,7 @@ def any_paths_are_invalid(
 
 @fail_function(include_if_all_elements_present=["results__df_with_mapper"])
 def paths_are_missing_in_targets_tree_mapper(
-    results__tree: NestedData,
+    results__tree: NestedResults,
```
MImmesberger, line 243: I don't get why we need to distinguish between internal data (`NestedData`) and result data (`NestedResults`). The type definition seems much more strict for `NestedData` than for `NestedResults`, so it can't be because internal data is not "clean".

src/ttsim/interface_dag_elements/automatically_added_functions.py
```diff
@@ -213,6 +230,13 @@ def _create_one_set_of_time_conversion_functions(
         if isinstance(element, ColumnFunction)
         else set()
     )
+    # `ScalarParam.start_date`/`end_date` are typed `date | None`, but every
+    # convertible element in the policy environment carries concrete dates
+    # (column objects and param functions always do; scalar params are built
+    # with explicit `start_date`/`end_date`). Pin the dates here so the
+    # `TimeConversionFunction` constructor receives non-optional values.
+    start_date = element.start_date or DEFAULT_START_DATE
```
MImmesberger: I don't quite get the comment, it is typed `date | None` but it never actually is `None`? Then we wouldn't need this here. Also, I thought we already did this via the policy function decorator.

src/ttsim/interface_dag_elements/automatically_added_functions.py
```diff
@@ -321,10 +368,24 @@ def create_agg_by_group_functions(
...
+            source_kind = _resolve_source_column_kind(
+                source_name=base_name_with_time_unit,
+                column_functions=column_functions,
+                qname_policy_environment=qname_policy_environment,
+            )
+            # Stamp concrete column-type annotations onto the `grouped_sum`
+            # wrapper. Its runtime implementation signature widens to
+            # `FloatColumn | IntColumn`; left untouched, that union becomes
+            # the node's producer type and the DAG's annotation-consistency
+            # check rejects it against a concretely typed consumer (Bug E).
```
MImmesberger: Bug E? 😄

src/ttsim/tt/column_objects_param_function.py
```diff
@@ -620,6 +722,71 @@ def _fail_if_other_arg_is_invalid(
...
+    typed `__signature__`. The primitive's runtime implementation signature
+    carries an imprecise `FloatColumn | IntColumn`-style union; resolving
+    the concrete output type from the source column's declared kind and the
+    aggregation rule table lets the aggregation node advertise an honest
+    producer type to the DAG's annotation-consistency check.
+
+    A `COUNT` aggregation always resolves (its output is an `IntColumn`
+    regardless of input). For the value aggregations, a decorated function
+    whose source-column parameter is unannotated keeps the primitive's
+    imprecise union — the user, not the build-time sweep, owns that node,
```
MImmesberger: (Maybe we should add to ai-instructions that it should avoid the word "sweep" at all costs, it's everywhere and it is never self-explanatory 😅)

src/ttsim/tt/column_objects_param_function.py
```diff
@@ -620,6 +722,71 @@ def _fail_if_other_arg_is_invalid(
...
+    mapper: dict[str, str],
+    agg_type: AggType,
+    func: Callable[..., Any],
+    column_name: str | None,
+    node_name: str,
+) -> Callable[..., Any]:
+    """Adapt an aggregation primitive to the DAG and stamp concrete types.
+
+    The primitive (`grouped_sum`, `sum_by_p_id`, …) is first renamed to the
+    DAG's argument names, then — when the source column's kind can be
```
MImmesberger, line 801: Maybe the heat is getting to my head but the second part of that sentence doesn't make sense, right?

src/ttsim/tt/param_objects.py
```diff
@@ -186,32 +212,44 @@ def __post_init__(self) -> None:
...
     _MIN_COEFFICIENTS_QUADRATIC = 2
     _MIN_COEFFICIENTS_CUBIC = 3
 
     @property
-    def slope(self) -> float:
-        """The first coefficient (linear term)."""
+    def slope(self) -> float | Float[Array, ""] | Int[Array, ""]:
+        """The first coefficient (linear term).
+
+        Indexing `coefficients` yields a 0-d array under the JAX backend and a
```
MImmesberger: This class is user-facing, so I would avoid this technical docstring here, also applies to the other methods.

src/ttsim/tt/aggregation_numpy.py
```diff
@@ -5,12 +5,13 @@
...
 
+from ttsim.typing import BoolColumn, DatetimeColumn, FloatColumn, IntColumn
+
 if TYPE_CHECKING:
-    from ttsim.typing import BoolColumn, FloatColumn, IntColumn
+    from ttsim.typing import BoolColumn, DatetimeColumn, FloatColumn, IntColumn
 
 
 def grouped_count(group_id: IntColumn) -> IntColumn:
-    fail_if_dtype_not_int(group_id, agg_func="grouped_count")
```
MImmesberger, line 13: Not sure whether these should go. Ofc, beartype will take care of this but many users may pass untyped policy functions to GETTSIM. IIUC there will be no checks for custom aggregators then.

Feel free to postpone until the GEP you alluded to is drafted.

src/ttsim/tt/type_resolution.py
```diff
@@ -0,0 +1,685 @@
+"""Build-time DAG type resolution for auto-generated nodes.
+
+ttsim auto-generates two kinds of DAG nodes: aggregation functions (`x_hh`
+from `x`) and auto-vectorized wrappers of scalar policy functions. Both are
+built by wrapping a primitive whose runtime signature carries an imprecise
+*union* return type (`grouped_sum` is statically `@overload`-ed per input
```
MImmesberger:
```suggestion
*union* return type (e.g. `grouped_sum` is statically `@overload`-ed per input
```

Not 100% sure

tests/interface_dag_elements/test_num_segments.py
```diff
@@ -1,7 +1,17 @@
 from __future__ import annotations
 
+from pathlib import Path
+
+import ttsim.interface_dag_elements
 from ttsim.interface_dag_elements.interface_node_objects import InterfaceFunction
-from ttsim.interface_dag_elements.num_segments import num_segments
+from ttsim.interface_dag_elements.orig_policy_objects import load_module
+
+# The package claw rewrites normally-imported `interface_dag_elements` modules,
```
MImmesberger, line 9: Users might hit the same issue with beartype on. Might make sense to provide a helper

src/ttsim/_jaxtyping_patch.py
MImmesberger, line 1: Didn't look into this one, IIUC this will be gone soon anyhow

src/ttsim/typing.py
```diff
@@ -1,17 +1,183 @@
...
+    NestedResults = Mapping[str, object]
+
+# User-boundary data aliases (Decision 8 / GEP-09). User-facing `InputData.*`
+# factories accept a wider leaf type than the canonical column aliases above:
+# users legitimately pass `pd.Series` and plain Python lists/sequences of
+# numbers, which internal code canonicalizes to backend arrays. Keep these
+# strictly at the `@beartype`-decorated user boundary; internal call sites use
+# the narrow `NestedData` / `FlatData` / `QNameData` forms.
+UserColumn: TypeAlias = (
+    FloatColumn | IntColumn | BoolColumn | pd.Series | list[float | int | bool]
```
MImmesberger: Should this be `UserFloatColumn` etc?

src/ttsim/tt/type_resolution.py
```diff
@@ -0,0 +1,685 @@
...
+
+    Returns:
+        The `ResolvedKind` for the annotation; `OTHER` for any annotation
+        that denotes neither a numeric column nor a numeric scalar.
+
+    Raises:
+        TypeResolutionError: If `annotation` is empty / missing — a node
+            that must be resolved has to carry an annotation.
+    """
+    if annotation is None or annotation == "" or annotation is _EMPTY:
```
MImmesberger, line 200: This effectively forces users to provide annotations, right? Not a blocker, might be a good idea to add an annotations tutorial though.

Maybe we want to enforce annotations for any function then? These special cases might be annoying.

MImmesberger (approved):
Cool, only some minor things.

It might make sense to release a new TTSIM version before merging this one. There might be issues we don't see right now that will surface once we use the new features in practice.

src/ttsim/plot/dag/interface.py
hmgaudecker: Not quite, we need it because we don't want to make Jax a strict dependency. See #106.

hmgaudecker:
> It might make sense to release a new TTSIM version before merging this one. 

Fair enough, on its way!

src/ttsim/interface_dag_elements/fail_if.py
hmgaudecker: The asymmetry is direction + DAG position, not strictness on a single scale. Flow:

1. **User boundary**: `UserNestedData` / `UserFlatData` / `UserQNameData` — wide leaves (pandas Series, Python scalars, JAX arrays, …).
2. **`_canonicalize_*` boundary**: narrows to typed backend columns.
3. **Internal DAG**: `NestedData` / `FlatData` / `QNameData` — strict, columns only. This is the contract the canonicalizer must produce.
4. **DAG output**: `NestedResults` / `FlatResults` / `QNameResults` — `object` leaves, because targets can be non-columns (param dicts, scalars, lookup arrays).

So the "strictness" is load-bearing: it pins what `_canonicalize_*` must emit, and what every internal function past that boundary can rely on. The asymmetry would be the wrong shape if "results" meant "computed columns" (your intuition) — but in our DAG it means "whatever the requested target evaluates to," which is genuinely heterogeneous.

Agreed the names mislead. I'd rather not bundle a rename into this PR (it'd touch every downstream typing import); if you find the wall worth dismantling, happy to do `NestedData → CanonicalNestedColumns` (or similar) as a follow-up.

src/ttsim/interface_dag_elements/tt_targets.py
hmgaudecker: **Renamed** to `QNameTTTargets` and `QNameColumnObjects` across ttsim + gettsim re-exports (14 files, ~46 occurrences). Convention is now clean: `Flat*` = tuple keys, `QName*` = str keys.

src/ttsim/plot/dag/interface.py
hmgaudecker: Still a good point -- I deleted the comment and the redundant `if TYPE_CHECKING:` re-import below it. Hoisting out of `TYPE_CHECKING` for the claw is the default, not a special case worth annotating per file. Applied the same cleanup to the three sibling sites you flagged.

src/ttsim/interface_dag_elements/automatically_added_functions.py
hmgaudecker: Indeed, the comment was defending dead code. Dropped both the `or DEFAULT_*_DATE` fallback and the comment. The assignment now reads the field directly with a `cast` for the type checker. The "every convertible element carries concrete dates" invariant is the source of truth.

src/ttsim/interface_dag_elements/automatically_added_functions.py
hmgaudecker: 😄 Stripped — that was a stray internal-tracker reference.

src/ttsim/tt/column_objects_param_function.py
hmgaudecker: Fair enough. https://github.com/OpenSourceEconomics/ai-instructions/pull/8

src/ttsim/tt/column_objects_param_function.py
hmgaudecker: It doesn't. Rewrote that branch to say what actually happens: "For value aggregations, a source-column parameter annotated with a non-numeric type leaves the primitive's imprecise union in place — without a numeric kind to narrow against, the build-time resolver ships the wider type rather than failing decoration." Left the code path itself (still reachable when the source-column annotation is, e.g., `str`).

src/ttsim/tt/param_objects.py
hmgaudecker: Trimmed. Dropped the "0-d under JAX, `numpy.float64` under NumPy" prose from all four spots in `PiecewisePolynomialInterval` (`.slope`, `.quadratic`, `.cubic`, plus the `ConsecutiveIntLookupTableParamValue.__call__` docstring that had the same issue). The return-type annotation already documents the variation; the prose was both user-unfriendly and redundant.

src/ttsim/tt/aggregation_numpy.py
hmgaudecker: Decided to address it now rather than postpone. Added universal annotation enforcement at decoration time for all five `@*_function` decorators (`@policy_function`, `@param_function`, `@agg_by_p_id_function`, `@agg_by_group_function`, `@group_creation_function`): missing parameter or return annotations raise the decorator's `*DefinitionError` immediately. This closes the "no checks for custom aggregators" gap you flagged — a user can't pass an untyped aggregator anymore. Audited ttsim, mettsim, gettsim, and gettsim-personas: zero existing breakage. GEP 9 + GEP 4 paragraphs describe this.

src/ttsim/tt/type_resolution.py
hmgaudecker: Applied verbatim.

tests/interface_dag_elements/test_num_segments.py
hmgaudecker: Held off on adding a public helper (would commit to a permanent debug API for a fairly niche case). Instead added an FAQ entry in `gettsim/docs/gettsim_developer/how-to-contribute.md` documenting the gotcha and the two workarounds: unset `TTSIM_BEARTYPE_CLAW`, or call `ttsim.interface_dag_elements.orig_policy_objects.load_module(path, root)` directly. Easy to promote to a public helper later if user demand surfaces.

src/ttsim/_jaxtyping_patch.py
hmgaudecker: Indeed, see #111.

src/ttsim/typing.py
hmgaudecker: Yes, that's cleaner. Recomposed to `UserColumn = UserFloatColumn | UserIntColumn | UserBoolColumn | list[float | int | bool]`. No semantic change (each `User*Column` already includes `pd.Series`), just makes the relationship to the per-kind aliases explicit.

src/ttsim/tt/type_resolution.py
hmgaudecker: Confirmed and answered both ways. Yes, the existing check already forces annotations for any function that reaches type resolution; this PR extends that to a universal contract at decoration time for all five `@*_function` decorators (`@policy_function`, `@param_function`, `@agg_by_p_id_function`, `@agg_by_group_function`, `@group_creation_function`). Every parameter and the return must carry an annotation, otherwise the decorator's `*DefinitionError` fires immediately with a stack trace pointing at the user's function definition. Much nicer than the dags' annotation errors.

Special cases drop out: there are no more "annotation-needed only on the auto-aggregation path" surprises, and the existing `_fail_if_annotations_mismatch_vectorization_strategy` check for `@policy_function` becomes a refinement on top of the universal precondition rather than the only gate.

Audited ttsim, mettsim, gettsim, and gettsim-personas before flipping the switch: 582 decorated functions, all already fully annotated, zero existing breakage. GEP 9 and GEP 4 reflect this universal rule now (on the `gep-09` branch). Annotations tutorial — I'd happily add one in a follow-up; pointer-only for now.

tests/interface_dag_elements/test_num_segments.py
MImmesberger: Ah, yes. This only affects `InterfaceFunction` objects and users will never write their own instances. Missed that.

MImmesberger (approved):
Looks good now, we can merge once the new TTSIM version is out.

PR #102, opened 2026-05-23
Canonicalize input dtypes; coerce uint, normalize nullable/pyarrow (closes #97, #94)

### LINKED ISSUE #97: BUG: Bug when inputs have uint dtype
### Bug description

TTSIM produces a (silent) bug when operating with uint dtypes. Often, variables are measured as uints (e.g. earnings which cannot be negative by definition). However, internally we do all sorts of operations with them  making the uint dtype invalid.

I stumbled over this issue when looking at the output of the following policy function:

```python
@policy_function()
def einnahmen_nach_abzug_werbungskosten_y(
    einnahmen__bruttolohn_y: float,
    werbungskosten_y: float,
) -> float:
    """Take gross wage and deduct Werbungskosten."""
    return max(einnahmen__bruttolohn_y - werbungskosten_y, 0.0)
```

Now, pyarrow would throw an `ArrowInvalid` error here. However, internally, we transform uint dtypes to numpy uints. **They produce silent bugs in these cases** 

```python
def test_uint_bruttolohn_does_not_overflow():
    result = max(np.uint32(0) - np.uint32(1230), 0.0)
    assert result == 0.0 # fails because result returns 4294966066.0
```

The following is a reproducer that demonstrated the bug still occurs if the input is `uint32[pyarrow]`:

```python
def test_uint_bruttolohn_does_not_overflow():
    result = main(
        main_target=MainTarget.results.tree,
        tt_targets=TTTargets.qname([
            "einkommensteuer__einkünfte__aus_nichtselbstständiger_arbeit__einnahmen_nach_abzug_werbungskosten_y",
        ]),
        input_data=InputData.tree({
            "p_id": pd.Series([1]),
            "einnahmen": {
                "bruttolohn_y": pd.Series([0], dtype="uint32[pyarrow]"),
            },
        }),
        policy_date_str="2023-01-01",
    )

    einnahmen = result["einkommensteuer"]["einkünfte"]["aus_nichtselbstständiger_arbeit"]["einnahmen_nach_abzug_werbungskosten_y"]
    assert float(einnahmen[0]) == 0.0
```

#### Proposed Solution

Transform uints internally to regular ints/floats.

Related to #94

hmgaudecker:
## Summary

A single ``_canonicalize_input_dtype`` helper in ``processed_data.py`` now normalizes every kind of non-canonical input column to numpy:

- **Unsigned integers** (numpy uint, pandas-nullable ``UInt*``, ``uint*[pyarrow]``) → ``int64`` (``int32`` on jax with x64 disabled) so signed arithmetic does not silently underflow into a huge positive value.
- **Float-nullable / ``float*[pyarrow]``** → ``float64`` with ``pd.NA`` mapped to ``NaN``.
- **Int / Bool nullable variants without NA** → ``int64`` / ``bool``.
- **Int / Bool nullable variants with NA** are surfaced by the new ``input_data_has_int_or_bool_missing_values`` fail-if, which reports every offending qname plus the first NA row in one message.

The existing ``input_data_uint64_values_overflow_int64`` fail-if continues to catch the only loss-of-precision case (``uint64 > int64.max``).

Closes #97. Closes #94.

Based on ``beartype`` (#101) — the ``ModuleType`` hoist in ``processed_data.py`` is needed under the package-wide claw.

## Test plan

- [x] ``pixi run -e py314 tests -n 7`` — 1022 passed.
- [x] ``pixi run -e py314-jax tests-jax -n 7`` — 1005 passed.
- [x] ``pixi run ty`` — no new errors in touched files.
- [x] Unit-parametrized over numpy uint{8,16,32,64}, pandas-nullable ``UInt32``, pandas-nullable ``Int*`` / ``Float*`` / ``boolean``, and ``int32/uint16/float64/bool[pyarrow]`` (auto-skipped when pyarrow isn't installed).
- [x] Integration test with mettsim's ``payroll_tax.income.amount_y`` mirroring the reproducer in #97.
- [x] Fail-if cases for ``uint64`` overflow + within-range pass-through + non-uint64 column ignore.
- [x] Fail-if cases for NA in single int column, NA in single bool column, and multi-column NA (all offending qnames reported).
- [x] End-to-end Float64-nullable + ``pd.NA`` round-trips as ``NaN``.


hmgaudecker (commented):
Autoreview.

src/ttsim/interface_dag_elements/data_converters.py
```diff
@@ -171,16 +176,19 @@ def df_with_nested_columns_to_flat_data(
...
     result = {}
-    for key, value in df.to_dict(orient="list").items():
-        clean_key = _remove_nan_from_keys(key)
-
-        # Use numpy for array creation if JAX backend is chosen and
-        # immediately convert back to JAX array.
-        # Performance optimization for JAX, PR #34
-        if backend == "jax":
-            result[clean_key] = xnp.asarray(numpy.asarray(value))
+    # `to_dict(orient='list')` strips nullable / pyarrow dtypes — read each
```
MImmesberger: Comment describes past behavior

src/ttsim/interface_dag_elements/fail_if.py
```diff
@@ -325,6 +325,84 @@ def input_data_is_invalid(input_data__flat: FlatData, xnp: ModuleType) -> None:
...
+            f"    - {dt.qname_from_tree_path(path)}: first NA at row {row}"
+            for path, row in offending
+        )
+        msg = format_errors_and_warnings(
+            "The following integer or boolean input columns contain missing "
+            "values, which cannot be represented in numpy integer / bool "
+            "dtypes:\n"
+            f"{formatted}\n\n"
+            "Fill the missing values, or convert the column to a float type "
+            "that supports NaN."
```
MImmesberger: This suggests users are allowed to pass NaN. We should not encourage this

src/ttsim/interface_dag_elements/fail_if.py
```diff
@@ -325,6 +325,84 @@ def input_data_is_invalid(input_data__flat: FlatData, xnp: ModuleType) -> None:
...
+def input_data_uint64_values_overflow_int64(input_data__flat: FlatData) -> None:
+    """Fail when a uint64 input column has any value outside the int64 range.
+
+    Uint columns are coerced to int64 in ``processed_data`` so that signed
+    arithmetic on them stays signed. Values in ``(int64.max, uint64.max]`` cannot
+    be represented as int64 and would silently wrap; surface them as an error.
+
+    Scope: only applies to raw ``numpy.uint64`` arrays supplied via
+    ``InputData.flat`` / ``InputData.tree`` / ``InputData.qname``. DataFrame
+    inputs go through ``_canonicalize_series`` first, which casts ``UInt64``
```
MImmesberger: Any reason why we don't do this for the other input shapes?

MImmesberger (approved):

src/ttsim/interface_dag_elements/data_converters.py
hmgaudecker: Rewritten to describe the current state: reading each column as a `pd.Series` preserves the original (possibly nullable / pyarrow) dtype for the canonicalizer.

src/ttsim/interface_dag_elements/fail_if.py
hmgaudecker: Dropped the "or convert the column to a float type that supports NaN" suggestion from both the docstring and the error message. 

The only encouraged action is now "Fill the missing values before passing the data."

src/ttsim/interface_dag_elements/fail_if.py
hmgaudecker: Good catch. `_canonicalize_input_dtype` (and `_canonicalize_series`) now run a uint-overflow check before the int64 cast, so DataFrame, `pd.Series`, and raw `numpy.uint*` inputs are all protected. An optional `column_label` is threaded through every call site (data_converters, processed_data, input_data) so the raised `ValueError` names the offending qname. The standalone `input_data_uint64_values_overflow_int64` fail-if is kept and its docstring updated to reflect that it now only catches raw `numpy.uint64` arrays that survive into `input_data__flat` with their dtype intact.

New test `test_uint64_overflow_from_dataframe_fails_with_actionable_message` pins the DataFrame-shape coverage end-to-end with a pandas-nullable `UInt64` column.

PR #103, opened 2026-05-23
Add symmetric qname I/O for InputData and results

hmgaudecker:
## Summary

Mirror `results.df_with_qname_columns` with the input-side counterpart so the qname format works in both directions.

- `InputData.qname({"a__b": ...})` becomes a live factory. Previously the factory existed in the public API but no interface function consumed `input_data__qname`, so calling `main(input_data=InputData.qname(...))` failed with a structural-validation error.
- `flat_from_qname` interface function converts the qname-keyed mapping to the tree-path-keyed flat form the rest of the pipeline expects. Each leaf goes through the existing dtype canonicaliser (so users can pass numpy arrays, pandas Series including nullable / pyarrow dtypes, or plain Python lists, exactly like the other factories).
- `MainTarget.input_data.qname` is added so `main(main_target=MainTarget.input_data.qname, ...)` resolves like the other entries.
- The `labels.input_columns` and `labels.empty_set` input-dependent interface functions learn about `input_data__qname` so auto-aggregation, the input-columns label, and the missing-root-nodes fail-if all behave symmetrically across factories.
- Output-side `results.df_with_qname_columns` returns a DataFrame whose flat string columns are the qualified names of the result paths — added in the original PR scope.

## Test plan

- [x] `pixi run -e py314 tests -n 7` — 1041 passed
- [x] `pixi run -e py314-jax tests-jax -n 7` — 1024 passed
- [x] `pixi run ty` — clean on touched files
- [x] `prek run --all-files` — clean
- [x] New end-to-end test variant exercises `InputData.qname` round-trip against the same fixture as the df-based factories.


hmgaudecker (commented):
Autoreview.

@MImmesberger, I hope this serves your use-case?

MImmesberger (commented):
We should make sure `Results` and `InputData` are symmetric. That is, we should add `qname` and `flat` to `Results` and `df_with_qname_columns` to `InputData`.

`df_and_mapper` vs. `df_with_mapper` difference is fine because both behave differently.

Do you agree?

hmgaudecker:
> We should make sure `Results` and `InputData` are symmetric. That is, we should add `qname` and `flat` to `Results` and `df_with_qname_columns` to `InputData`.
> 
> `df_and_mapper` vs. `df_with_mapper` difference is fine because both behave differently.
> 
> Do you agree?

Agreed -- better now?

MImmesberger (approved):
Yes, thanks!

PR #104, opened 2026-05-23
Partial scalar inputs that drive derived time-conversion functions (#79)

### LINKED ISSUE #79: Partialling errors of derived inputs in tt_function

This issue tracks the partialling error of inputs that would be derived functions in the `tt_function`. Here's an example:

- The `policy_input` is `x_m`
- You pass `x_y` as a scalar when requesting the `tt_function`
- Then, the `tt_function` will still ask for `x_y` as an input, i.e. the partialling that should have happened in step 2 did not happen

The reason is connected to the problems we had when building templates: Derived functions are not created when there is no input data to create them from. In particular, this line in `with_processed_params_and_scalars` is problematic, because there is no node `x_y` when creating the `tt_function` in the example above:

```python
    for n, f in without_tree_logic_and_with_derived_functions.items():
        if n in processed_data:
             ... # This is where the partialling is prepared
```

Similarly, partialling doesn't work with inputs of already vectorized functions.

_Originally raised in PR #54._

hmgaudecker:
## Summary
- Closes #79.
- ``with_processed_params_and_scalars`` (``specialized_environment.py:152``) used to only insert scalars from ``processed_data`` when the qname was already a key in ``without_tree_logic_and_with_derived_functions``. The time-conversion machinery builds conversions *away from* a source qname (e.g. ``x_m``/``x_w``/``x_d`` from a yearly ``x_y``) but never creates a self-node for the source. So a scalar ``x_y`` lived in ``processed_data`` yet never appeared in the env; the derived ``x_m = per_y_to_per_m(x_y)`` had no ``x_y`` to partial in and surfaced as an unbound root node of the tt_dag — exactly the symptom reported in #79.
- Patch is a 9-line second pass after the existing loop: for every qname in ``processed_data`` not yet in ``all_nodes``, if its value is a scalar, lift it into the env. The downstream ``with_partialled_params_and_scalars`` then sees the scalar and bakes it into its consumer.
- Two pre-existing xfail tests pinning this exact bug (``test_specialized_environment.py:885``, ``test_end_to_end.py:285``) now pass — strict-xfail markers removed.

## Scope
Kept narrow per #79: the partialling now works internally to the env. End-to-end paths that pass plain scalars to natively-vectorized consumers continue to hit ``fail_if.passed_scalar_inputs_for_natively_vectorized_functions`` with its clearer ``"must be passed as arrays"`` message; relaxing that broader guard rail is a separate policy decision.

## Test plan
- [x] ``pixi run -e py314 tests -n 7`` — 971 passed.
- [x] ``pixi run -e py314-jax tests-jax -n 7`` — 954 passed.
- [x] ``pixi run ty`` / ``pixi run ty-jax`` — clean.
- [x] The two previously-xfail regression tests now pass on both backends.


hmgaudecker (commented):
Autoreview.

@MImmesberger, if you still have the original use case floating around, would be great to double-check. But the tests may be good enough.

MImmesberger (approved):
I don't have the original issue here, but the tests are good enough IMO

PR #106, opened 2026-05-23
Document JAX-optional constraint on typing module (#84)

### LINKED ISSUE #84: Explanation: Keep `TYPE_CHECKING` and `from __future__ import annotations` to avoid `jaxtyping.Array` as dependency

We explored removing `TYPE_CHECKING` blocks and `from __future__ import annotations` to simplify typing and imports. This turns out not to be viable because we want to avoid JAX as a runtime dependency.

- `Bool`, `Float`, and `Int` from `jaxtyping` can be imported at runtime without pulling in JAX.
- `jaxtyping.Array` cannot:
  - Importing `Array` triggers a JAX import via a `__getattr__` hook in `jaxtyping.__init__`.

Column types are defined as `IntColumn = Int[Array, "n_obs"]`. Evaluating these aliases at runtime would therefore implicitly import JAX.

So, for now:
- Column type aliases of the form `<DType>[Array, "<shape>"]` must not be evaluated at runtime.
- Imports required for defining these aliases must remain inside `TYPE_CHECKING`.
- The column registry must continue returning string annotations.
- `from __future__ import annotations` remains required in files that reference these column type aliases.

hmgaudecker:
## Summary
- Closes #84.
- Expands the module docstring in ``src/ttsim/typing.py`` to capture why ``TYPE_CHECKING`` + ``from __future__ import annotations`` must stay in place across the codebase: ``jaxtyping.Array`` resolves via a top-level ``__getattr__`` hook that imports JAX, so any module touching ``Array`` at module scope would force JAX in as a runtime dependency.
- The new docstring is structured as bottom line → mechanism → do's-and-don'ts bullets for maintainers.
- Cross-reference to the shared ``ai-instructions`` submodule deferred (submodule lives in a separate OSE-wide repo); this PR is project-local.

## Test plan
- [x] ``pixi run ty`` — clean.
- [x] ``prek run --files src/ttsim/typing.py`` — clean.
- Docs-only change; no behavioural tests.


src/ttsim/typing.py
```diff
@@ -1,16 +1,51 @@
 """Type aliases used across ttsim.
 
+Bottom line: JAX must remain an *optional* runtime dependency of ttsim.
```
hmgaudecker:
```suggestion
Bottom line: JAX shall remain an *optional* runtime dependency of ttsim.
```

MImmesberger (approved):
Not sure whether this should go in the typing module or rather in the developer docs (or docs exclusively used by AI).

PR #107, opened 2026-05-23
Broadcast scalar leaves in InputData.tree to length-n_obs (#60)

### LINKED ISSUE #60: BUG: Treatment of scalars in input data varies by input format

When users provide a dataframe and a mapper, scalar values are broadcast to arrays matching the length of `p_id`. But when users provide the exact same data as an input tree, we do **not** broadcast; instead, we partially apply the scalars inside the TT function.

#### Describe the solution you would like to see

Since the partial-application feature does not work well yet, we should keep the current behavior of the dataframe + mapper pipeline; otherwise, casual users will run into errors that are nearly impossible to debug.

So either:

1. we find a way to fix the partial-application issue for all inputs, or  
2. **(preferred in the short run)** broadcast scalar inputs for input trees as well, and require users who want to rely on partial-application to supply their inputs via `processed_data` instead.

hmgaudecker:
## Summary
- Closes #60.
- ``flat_from_tree`` in ``input_data.py`` now broadcasts scalar leaves to length-``n_obs`` arrays (``n_obs`` derived from ``p_id``). The two df-based paths already broadcast; ``InputData.tree`` now matches them so the cross-path equivalence promised by #60 holds.
- ``InputData.flat`` remains untouched and stays the opt-out for users who rely on scalar partial-application (introduced in #79). ``UserFlatData`` / ``UserQNameData`` widen to accept scalar leaves so the opt-out keeps a working type signature.

## Test plan
- [x] New ``tests/interface_dag_elements/test_input_data.py`` with three tests: tracer-bullet broadcast, cross-path equivalence across ``InputData.{tree, df_and_mapper, df_with_nested_columns}``, and opt-out preservation via ``InputData.flat``.
- [x] Existing tests that pinned partial-application on the tree path were updated to use ``InputData.flat`` (the opt-out where partialling still applies). The scalar-feeding-natively-vectorized fail-if stays in place for that path.
- [x] ``pixi run -e py314 tests -n 7`` — 1022 passed.
- [x] ``pixi run -e py314-jax tests-jax -n 7`` — 1005 passed.
- [x] ``pixi run ty`` / ``pixi run ty-jax`` — clean.


hmgaudecker (commented):
Autoreview.

@MImmesberger, is this about what you had in mind how things should work?

MImmesberger:
I would like to push back on this. I think what we really want is that users can pass scalars for all input columns and these get partialled into the TT function. If we managed to get this right, issue #60 is fixed automatically.

I have drafted an alternative here: #113. We should discuss whether this implementation is future-proof and which direction we want to go.

hmgaudecker:
Superseded by #113. The lazy-broadcast design preserves scalar partialling for auto-vectorized functions and removes the 'scalar-fed-natively-vectorized' fail-if. See #113 for the rationale and implementation.

PR #108, opened 2026-05-23
Reverse-translate endogenous p_id_* targets to user p_ids (closes #43)

### LINKED ISSUE #43: ENH: Allow returning endogenously-computed pointers to `p_id`
When exogenously-given pointers to another `p_id` are requested as targets (say, a user wants to have `p_id_parent_1` to be part of her results to save a merge operation), we just return `p_id_parent_1` from the input data directly.

When computing pointers endogenously, we only know the internal `p_id`. Because of that, we are raising `fail_if__endogenous_p_id_among_targets` as of #42.

This should not be hard to implement, but without a use case it seems unnecessary.

hmgaudecker:
## Summary
- `raw_results.columns` now reverse-translates any output column whose leaf name starts with `p_id_` from internal indices back to user-space `p_id` values, preserving the `-1` no-link sentinel.
- Removes the previously protective `fail_if.endogenous_p_id_among_targets` (and its `MainTarget` entry / tests) — it's no longer needed.
- Four new end-to-end tests cover the tracer-bullet case, `-1` preservation, exogenous (input-data) `p_id_*` passthrough, and the mixed-with-regular-column case.

Closes #43.

## Test plan
- [x] `pixi run -e py314 tests -n 7` — 1023 passed
- [x] `pixi run -e py314-jax tests-jax -n 7` — 1006 passed
- [x] `pixi run ty` — no new errors in touched files
- [x] `prek run --all-files` — clean


hmgaudecker (commented):
Autoreview.

src/ttsim/interface_dag_elements/raw_results.py
```diff
@@ -20,15 +22,59 @@ def columns(
...
     """
-    return tt_function(
+    raw = tt_function(
         {k: v for k, v in processed_data.items() if k in labels__root_nodes},
     )
+    sorted_orig_p_ids = xnp.asarray(input_data__flat[("p_id",)])[
+        input_data__sort_indices
+    ]
+    return {
+        qname: _translate_internal_to_orig_p_ids(value, sorted_orig_p_ids, xnp)
```
MImmesberger: Use keyword argument names in function call

MImmesberger (approved):

src/ttsim/interface_dag_elements/raw_results.py
hmgaudecker: Done.

PR #111, opened 2026-05-24
Drop local jaxtyping sentinel-pickle patch in favour of upstream fix

hmgaudecker:
## Summary

Replace the local monkey-patch in `src/ttsim/_jaxtyping_patch.py` for the variadic-dim sentinel with an upstream jaxtyping fix that covers all three affected sentinels (`_any_dtype`, `_anonymous_dim`, `_anonymous_variadic_dim`), not just the one we were patching.

Until the upstream fix is released as `jaxtyping >= 0.3.10` on PyPI, this PR pulls the patched fork branch via a pixi `pypi-dependencies` git override — same pattern as the existing `dags` override. Once the release lands, drop the override; the floor pin `jaxtyping>=0.3.10` in `[project].dependencies` already pins consumers of the released wheel to the fixed version.

Mirrors [OpenSourceEconomics/pylcm#365](https://github.com/OpenSourceEconomics/pylcm/pull/365), which did the same swap for pylcm.

## Changes

- `pyproject.toml`: bump `jaxtyping>=0.2 → jaxtyping>=0.3.10`; add the pixi pypi-dependency override pointing at `hmgaudecker/jaxtyping@fix/sentinel-cloudpickle`.
- `src/ttsim/_jaxtyping_patch.py`: drop the `_AnonymousVariadicDim` singleton and its install. **Keep** the `Ellipsis ↔ "..."` shim — that one stays because `from __future__ import annotations` is still in use (ttsim's floor is Python 3.11), so stringified `Int[Array, ...]` annotations still `eval` back to `Int[Array, Ellipsis]` and need coercion before reaching jaxtyping's subscript handler. (Drop the shim once the floor reaches Python 3.14.)
- `src/ttsim/__init__.py`: update the import-comment to reflect the surviving purpose.
- `pixi.lock`: regenerated for all envs (`jaxtyping 0.3.9 → 0.3.10` from the fork).
- `CHANGES.md`: entry under Unreleased.

## Background

jaxtyping's three module-level `object()` sentinels do not survive a cloudpickle round-trip — the receiving side gets fresh `object()` instances that no longer match the live module-global, so identity-based shape checks on cloudpickled jaxtyping types fail. ttsim hits the variadic-dim variant through DAG-built `cloudpickle`-able callables under the beartype claw; the other two would fail similarly through different code paths.

Upstream issue: [patrick-kidger/jaxtyping#390](https://github.com/patrick-kidger/jaxtyping/issues/390). Fix branch: [`hmgaudecker/jaxtyping@fix/sentinel-cloudpickle`](https://github.com/hmgaudecker/jaxtyping/tree/fix/sentinel-cloudpickle) — replaces all three sentinels with `__reduce__`-backed singleton classes; upstream PR pending.

If jaxtyping 0.3.10 ships to PyPI before this PR merges, swap the override out in the same diff.

## Test plan

- [x] `pixi run --locked ty` (in `ttsim/`) — clean.
- [x] `pixi run --locked ty-jax` (in `ttsim/`) — clean.
- [x] `prek run --all-files` — clean.
- [x] `pixi run -e py314 tests ttsim/tests/` — **993 passed, 6 skipped, 2 xfailed**.
- [x] `pixi run -e py314-jax tests-jax ttsim/tests/` — **976 passed, 23 skipped, 2 xfailed**.

## Stacking

Stacked on #101 (beartype). The cascade through #102–#110 will pick this up via the regular merge-forward flow once this lands on `beartype`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,12 @@ releases are available on [Anaconda.org](https://anaconda.org/conda-forge/ttsim)
 
 ## Unreleased
 
+- Replace the local jaxtyping sentinel-pickle monkey-patch with an upstream fix that
+  covers all three `_array_types` sentinels (`_any_dtype`, `_anonymous_dim`,
+  `_anonymous_variadic_dim`). Bumps the floor pin to `jaxtyping >= 0.3.10`; pulls the
+  fork branch `hmgaudecker/jaxtyping@fix/sentinel-cloudpickle` via pixi until the
+  release lands. The `Ellipsis` ↔ `"..."` shim stays because
+  `from __future__ import annotations` is still in use. ({ghuser}`hmgaudecker`)
 - {gh}`101` Adopt the package-wide beartype claw with a typed exception hierarchy at
   user-facing boundaries (`InputDataError`, `EntryPointError`, `TTTargetsError`,
   `PolicyFunctionDefinitionError`, …) so malformed input is rejected with curated
--- a/pixi.lock
+++ b/pixi.lock
@@ -263,6 +263,7 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/noarch/widgetsnbextension-4.0.15-pyhd8ed1ab_0.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/zipp-3.23.0-pyhcf101f3_1.conda
       - pypi: .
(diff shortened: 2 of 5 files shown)


hmgaudecker:
Folded into #101 (commit b73b39ec) — jaxtyping 0.3.10 is now on PyPI, so the floor pin alone is sufficient and the git override / fork branch is unnecessary.

PR #113, opened 2026-05-25
Defer scalar→array broadcast to call time for non-vectorized functions

MImmesberger:
## What

Scalar inputs feeding functions that are **not** auto-vectorized — aggregations, group-creation functions, and `vectorization_strategy="not_required"` policy functions — are now broadcast to length `n_obs` at **call time**, instead of raising a fail-fast error. Auto-vectorized functions are unaffected: their scalars stay scalars (partialled in).

## `num_segments` → `len_p_id`

The `num_segments` interface node always returned `len(p_id)` (== `n_obs`), so it now does double duty as the broadcast length. Renamed it to `len_p_id` to reflect its content. The jax aggregation machinery still consumes `num_segments` (its correct term), fed from `len_p_id` via a bridge. `len_p_id` is partialled in as a scalar, so the broadcast introduces **no new DAG dependency**.

## Alternative

Instead of sizing the broadcast off the partialled `len_p_id`, we could make `p_id` an **implicit dependency of every non-auto-vectorized function** and read `n_obs` off it.

## Relation to #107

This is an **alternative to #107**, which broadcasts scalar leaves eagerly in `InputData.tree`. Here, scalars stay scalars and are only broadcast at call time, where a non-vectorized function actually needs a full-length array.


hmgaudecker:
Much nicer!

hmgaudecker (approved):
I'd be happy with this, but please double-check what I did! 

Thanks for this one!!!

MImmesberger (commented):
Looks good to me 👍

PR #122, opened 2026-06-03
pint units (GEP 10): framework core + time-as-a-dimension model

### LINKED ISSUE #117: pint units: framework core + tracer bullet
## What to build

Establish the pint-based unit framework in ttsim and prove it end-to-end on a single mettsim case. A module-level pint UnitRegistry defines the closed dimension vocabulary ([currency], [time], [area], [count], plus quarter and dimensionless). A register_currency(name, *, base=False, definition=None) API lets downstream packages add concrete currencies. The @policy_function decorator accepts a unit= string (the non-time part; the token CURRENCY denotes the [currency] dimensionality). A minimal per-function dry-run wraps a function inputs in representative Quantities of their declared units, executes the scalar body in numpy+pint, and infers the output unit. An always-on fail_if (developed against crafted fixtures at this stage) raises on a dimensionally invalid operation. Demonstrate on one mettsim function: a compatible case passes; a hand-made CURRENCY + CURRENCY/area clash fails with a clear error.

pint runs only at build time and never wraps a runtime array.

## Acceptance criteria

- [ ] pint UnitRegistry with [currency]/[time]/[area]/[count]/dimensionless + quarter
- [ ] register_currency API; mettsim registers its own base currency
- [ ] @policy_function accepts unit=, with the CURRENCY token checked at dimensionality level
- [ ] infer_function_unit(function, input_units) returns the inferred unit or raises on clash
- [ ] one mettsim function demonstrates both pass and fail
- [ ] new unit tests green; no runtime array is ever wrapped in a Quantity

## Blocked by

None - can start immediately

MImmesberger:
Closes #117. First PR of the GEP 10 (units and dimensionality) series.

## What this does

Establishes the pint-based unit framework and proves it end-to-end on a single mettsim function.

- **Closed unit vocabulary as a `Unit` enum** in a new `ttsim.tt.units` module. A declaration is one member of a closed, UPPER_SNAKE `Unit` enum (YAML uses the identical string, so the JSON schema enumerates the legal values). The members are **kind tokens**, period-abstracted: `CURRENCY_STOCK`, `CURRENCY_FLOW`, `SHARE_FLOW`, `YEARS`, `HOURS_FLOW`, `SQUARE_METERS`, `HECTARES`, `CURRENCY_PER_SQUARE_METER_FLOW`. Each maps internally to a (pint expression, kind) pair; pint strings are an internal representation only, never written in declarations.
  - The `[currency]` dimension is anchored by a real `CURRENCY` pint unit, so checks compare at the dimensionality level independent of import order / before any concrete currency is registered.
  - A dimensionless quantity (a share, a rate, a head count) declares **no unit**: `unit=None` in code, `unit: null` in YAML. Counts are dimensionless — there is no `[count]` dimension. The spellings `"dimensionless"`/`"count"` are rejected, so there is exactly one way to write it. An *omitted* `unit=` is the `UNSET_UNIT` sentinel (distinct from a `None` declaration), which the mandatory-units check reports in #124.
  - pint's built-in `quarter` is a unit of mass, so the time quarter is `quarter_year = year / 4`.
- **`register_currency(name, *, base=False, definition=None)`** — downstream packages register concrete currencies; mettsim registers **castar** (Gondor) as its base on import. Registration is idempotent, rejects a second base, and **derives the currency's declaration tokens** — one concrete variant per currency-dimensioned core token (`CASTAR_STOCK`, `CASTAR_FLOW`, `CASTAR_PER_SQUARE_METER_FLOW`, …).
- **Currency as a union.** The agnostic `CURRENCY_*` tokens denote the union of the registered currencies. Columns and functions accept only the agnostic `Unit` members and are thereby provably currency-agnostic; parameters use the concrete tokens to name the currency their numbers are written in (consumed by the conversion machinery in #125). For every dimensionality check a concrete token resolves to its agnostic counterpart — the checks never see a concrete currency.
- **`@policy_function` / `@policy_input` / `@param_function` accept `unit=`** (a `Unit` member, `None`, or unset). Validated at decoration time by the beartype claw — a bare string is rejected. The unit is threaded through every `remove_tree_logic` / `vectorize` reconstructor.
- **The build-time dry-run** wraps inputs in representative `Quantity`s, executes the scalar body in numpy+pint, and infers the output unit or raises on a dimensional clash. The per-function body check (inferred output must match the declared unit) is exercised here against crafted fixtures and one mettsim function; wiring it into the assembled environment as an always-on check is #126.

**pint runs only at build time and never wraps a runtime array** — the numeric/JAX path is untouched.

## Tracer bullet

`tests/tt/test_units.py`:
- `acre_size_in_hectares_after_cap` (mettsim) passes the per-function check (area in → area out).
- A hand-made `CURRENCY + CURRENCY/area` clash fails loudly.

## Acceptance criteria

- [x] pint registry with `[currency]`/`[time]`/`[area]` + `quarter_year`; closed `Unit` token vocabulary (unknown tokens rejected; dimensionless = `unit=None` / `unit: null`; counts dimensionless)
- [x] `register_currency` API deriving per-currency declaration tokens; mettsim registers castar as base
- [x] decorators accept `unit=`; currency compared at the dimensionality level (union semantics)
- [x] build-time dry-run returns the inferred unit or raises on clash
- [x] one mettsim function demonstrates both pass and fail
- [x] new unit tests green; no runtime array is ever wrapped in a Quantity

## Verification

All CI green (tests on py311–py314 incl. jax across Linux/macOS/Windows; ty; codecov). pint never wraps a runtime array.


src/ttsim/tt/units.py
```diff
@@ -0,0 +1,1213 @@
...
+    token. Replaces the former ``unit=None`` / ``unit: null`` spelling."""
+
+    DIMENSIONLESS_FLOW = "DIMENSIONLESS_FLOW"
+    """A dimensionless quantity *per period* (``1/period``): a count or a
+    share per unit time — births per year, or the per-year change of a
+    dimensionless factor (the pension Zugangsfaktor). The complete,
+    non-period counterpart is :attr:`DIMENSIONLESS`."""
+
+    YEARS = "YEARS"
+    """A quantity measured in years: ages, age thresholds, calendar years."""
```
hmgaudecker: **Calendar years and durations are different objects.** Extending `YEARS` to 'calendar years' conflates an affine point/ordinal with a duration. Real gettsim pension code mixes `y_to_m(geburtsjahr)` with a dimensionless month ordinal; under the dry-run that is either falsely rejected (`year + dimensionless`) or goes dimensionless and is unchecked. A distinct calendar-point/ordinal type or a unit-aware date-difference helper would resolve it. (See ttsim-dev/gettsim#1193, gep-10.md L164.)

src/ttsim/tt/units.py
```diff
@@ -0,0 +1,1213 @@
...
+        )
+    _fail_if_currency_token_collides(name)
+    if base and _base_currency not in (None, name):
+        raise UnitDefinitionError(
+            f"Cannot register {name!r} as the base currency: "
+            f"{_base_currency!r} is already the base currency."
+        )
+
+    currency_dim = UNIT_REGISTRY.Quantity(1.0, CURRENCY_TOKEN).dimensionality
+    if name in UNIT_REGISTRY:
```
hmgaudecker, line 596: **Re-registering a currency only checks dimensionality, not the conversion factor.** Re-registering `BASE/4` as `BASE/5` is accepted and silently keeps `/4`, and a derived currency can be re-registered `base=True` even though its factor against `CURRENCY` is 0.25 — either invalidates every later conversion factor. Comparing the requested vs existing factor, and rejecting `base=True` unless the existing unit is exactly factor 1 against `CURRENCY`, would close it.

hmgaudecker (commented):

src/ttsim/tt/units.py
```diff
@@ -0,0 +1,1213 @@
...
+        return arr.to(expected_unit).magnitude
+    except pint.DimensionalityError as e:
+        raise UnitConsistencyError(
+            f"pint-tagged input{where} has unit '{arr.units}', which is not "
+            f"compatible with the declared unit '{expected_unit}' (in the run "
+            f"currency)."
+        ) from e
+
+
+def strip_input_quantity_at_boundary(
```
hmgaudecker, line 1325: Cross-reference: this is the boundary strip that converts currency only and leaves area/time scale untouched — the other half of the same-dimension input mis-scaling reported on #125 (https://github.com/ttsim-dev/ttsim/pull/125#discussion_r3448675455).

hmgaudecker (commented):
Cross-reference to a GEP-10 audit finding.

src/ttsim/tt/units.py
MImmesberger: Fixed across the stack (pint-117 engine → pint-119 dry-run → pint-121 mettsim worked example). Added a **calendar-point sub-model**: six tokens in two kinds — `CALENDAR_YEAR/MONTH/DAY` (affine pint offset units = points on the calendar) alongside `YEARS/MONTHS/DAYS` (durations).

pint enforces the algebra: `point − point → duration`, `point ± duration → point`, while `point + point`, scaling a point, and mixing calendar axes raise `OffsetUnitCalculusError` — caught in the dry-run with a dedicated calendar-misuse message. `YEARS` now resolves to `delta_calendar_year` (ratio 1 vs `year`), so every existing `YEARS` annotation is unchanged.

The framework date nodes (`policy_year`, …) are now calendar points; cyclic ordinals like `geburtsmonat` stay `DIMENSIONLESS`. mettsim worked example: `geburtsjahr` (`CALENDAR_YEAR`) + computed `age = policy_year − geburtsjahr` (`YEARS`).

src/ttsim/tt/units.py
MImmesberger: Fixed in `25eda9ca` (pint-117). `register_currency` now compares the requested conversion factor against the existing one on re-registration, not just the dimensionality: `existing = Quantity(1.0, name).to(CURRENCY).magnitude` versus `1.0` for `base=True` or `parse_expression(definition).to(CURRENCY).magnitude` for a definition, rejected unless `math.isclose(existing, requested, rel_tol=_REL_TOL)`. This catches both halves of the finding — `castar / 4` re-registered as `castar / 5` (previously the old factor was silently kept) and re-registering the base with a divergent definition — while idempotent same-factor re-registration still passes. Regression test: `test_register_currency_with_inconsistent_factor_fails`.

src/ttsim/tt/units.py
MImmesberger: Addressed via the fail-node equivalence check in `1001a608` (see the detailed reply on #125). The live strip path stays currency-only by design; the upstream `fail_if_input_units_are_inconsistent` now guarantees the non-currency axes are ratio-1 equivalent to the declared unit before the strip runs, so leaving area / intrinsic-time untouched at the boundary is correct rather than a silent mis-scale.

MImmesberger:

PR #123, opened 2026-06-03
pint units: full dimension model — time as a dimension (#118)

MImmesberger:
Closes #118. Second PR of the GEP 10 series. **Stacked on #122 (#117)** — review/merge that first; the base retargets to `main` once #117 lands.

## What this does

Makes time a first-class pint dimension and routes the time-conversion factors through pint.

### pint-sourced conversion factors
`ttsim/unit_converters.py`'s four period-per-year constants are now derived from pint rather than hand-written:

```python
_Q_PER_Y = _as_int_if_integral(_ratio("year", "quarter_year"))  # 4  (int)
_M_PER_Y = _as_int_if_integral(_ratio("year", "month"))         # 12 (int)
_D_PER_Y = _ratio("year", "day")                                # 365.25
_W_PER_Y = _D_PER_Y / _ratio("week", "day")                     # 365.25 / 7
```

- The suffix auto-generation, GEP-1 naming, the ~50 converter functions and their stock/flow duality are **unchanged**.
- Integral factors stay `int` so stock converters keep integer inputs integer (`y_to_m(2) == 24`).
- The week factor is composed via days so it reproduces GEP 1's canonical `365.25 / 7` bit-for-bit (pint's direct `year / week` differs by one ULP).

### Token + period → full unit
New resolvers in `ttsim.tt.units` combine a kind token with its flow period:

- `resolve_column_unit(Unit.CURRENCY_FLOW, suffix="m")` → `CURRENCY / month`; the auto-generated `_y` variant → `CURRENCY / year`. A flow token takes its period from the name suffix (columns/functions).
- `resolve_param_unit(Unit.CURRENCY_FLOW, reference_period="Year")` → `CURRENCY / year` (`reference_period` is functional for parameters).
- A complete token (`CURRENCY_STOCK`, `YEARS`, …) is the whole unit and admits no period source; intrinsically temporal quantities (an age, a duration) use `Unit.YEARS` directly. A time suffix on a complete-token node, or a `…_FLOW` token with no period source, is an error.

**Currency tokens follow union semantics**: a concrete currency token resolves exactly like its agnostic counterpart for every check — the dimensionality pass never sees the concrete currency; it only drives the build-time conversion. `resolve_column_unit` rejects concrete currency tokens: columns and functions are currency-agnostic by design.

`[area]` and dimensionless quantities (`unit=None` / `unit: null`) are handled by the closed vocabulary from #122.

## Acceptance criteria

[...]

Issue #119, opened 2026-06-03
pint units: mandatory units + edge-consistency check


MImmesberger:
## What to build

Turn the dry-run into a complete, data-independent check. Make return units mandatory on every column/param/function decorator, raising a definition-time error (paralleling GEP-9 return-type enforcement) when missing. Add edge-consistency: across every DAG edge the producer declared unit must equal the consumer declared expectation. Auto-generated nodes (time variants, aggregations) receive auto-assigned units (SUM/MEAN/MIN/MAX preserve, COUNT to [count], ANY/ALL to dimensionless). Exercise against crafted fixtures and partially-annotated mettsim subtrees.

## Acceptance criteria

- [ ] missing unit raises at decoration time with a clear error
- [ ] per-function body check + edge-consistency check, both data-independent
- [ ] auto-generated nodes get correct auto-assigned units
- [ ] strict literal handling documented (bare literals in mixed-unit arithmetic must be unit-tagged)
- [ ] tests green

## Blocked by

- https://github.com/ttsim-dev/ttsim/issues/118

MImmesberger:
Closed via #138

PR #124, opened 2026-06-03
pint units (GEP 10): mandatory units + edge-consistency checks

MImmesberger:
Implements #119, third PR of the GEP-10 stack (on top of #123 → #122).

## What this does

Turns the dry-run into a complete, data-independent check (GEP 10) and ships the full environment-level check machinery (wired always-on in #126).

### Strict engine

- **`fail_if_units_are_missing`** flags any node without a `unit=` declaration — mandatory units, paralleling GEP-9 return-type enforcement. `unit=None` / `unit: null` is a dimensionless *declaration*, not a gap; an omitted `unit=` is the `UNSET_UNIT` sentinel, which is what this check reports.
- **`fail_if_unit_edges_are_inconsistent`** is the per-function body + edge check: each function's inputs are wrapped in the resolved units of their producers, the body is dry-run in numpy+pint, and the inferred output must match the node's own resolved unit. Feeding producer units into each consumer is the edge-consistency pass — a monthly producer reaching a yearly consumer is caught. Fully data-independent.

### Auto-assigned units on derived nodes

- `unit_for_aggregation`: SUM/MEAN/MIN/MAX preserve the source unit; COUNT/ANY/ALL → dimensionless. Wired into auto-generated time-conversion variants (which inherit the source token **verbatim** — a flow token is period-invariant, the variant reads its period off its own suffix) and into auto-aggregations; an auto-SUM over a **boolean** source resolves to dimensionless (a head count).
- Auto-aggregation resolution walks to the sibling-time-unit `PolicyInput` when the source column is supplied at a different time unit than its declaration.

### Decorators

- `@policy_input` validates `unit=` at decoration time like the other decorators.
- `@agg_by_group_function` / `@agg_by_p_id_function` accept `unit=`; COUNT/ANY/ALL auto-assign theirs.

### Environment-level check machinery (`interface_dag_elements/unit_checks.py`)

Resolves the complete unit of every node in an assembled environment (suffix-resolved for columns/param functions, `reference_period`-resolved for params) and wraps the strict engine with the conservative rules that make an always-on check safe:

- Structural exemptions: identifiers (`p_id`, `*_id`, `p_id_*`), `@group_creation_function`s, and **boolean nodes** (dimensionless by construction — never wrapped in a Quantity; any unit on a boolean node is rejected), plus framework date nodes.
- A mismatch is flagged only when the inferred unit is concrete and disagrees with the declaration; bodies that cannot be dry-run (piecewise, lookup tables, `join`/`xnp` ops) and dimensionless inferences fall back to the declaration.
- Bodies with boolean inputs are dry-run **twice** (all-truthy / all-falsy), covering both arms of `if exempt: return 0.0 else: <real arithmetic>` guards — the falsy run is what catches a stock×rate body missing its time component.

### Parameter metadata

- `ParamObject.unit` carries a `Unit` token (coerced from the YAML string in `__post_init__`); `params-schema.json` gains a `unitToken` enum (the legal tokens, restoring pre-commit validation) — the old free-text `"Euros"/"Share"/…` label enum is gone.
- **Dict parameters may declare per-leaf units** (`unit:` as a mapping leaf → token; `reference_period` must then be null). Dict params enter the dry-run as dicts of representative Quantities, so subscripting bodies are verifiable.
- **Function-like parameters declare `input_unit:` / `output_unit:` per axis** (piecewise schedules, lookup tables, phase-in/out) — a schedule is a function between quantities, not a quantity; a `unit:` on them is an error. Each axis token follows the kind rules (a `…_FLOW` axis takes its period from the single `reference_period`; a dangling `reference_period` is an error; the name-suffix coincidence applies to the **output** axis). The pair also tells the build-time conversion which numbers to rescale (#125). The schema enforces the split per `type:`.

### Currency strictness

Once a concrete currency is registered, an agnostic `CURRENCY_*` token on a **parameter** is an error — it must name its denomination (`SILVER_PENNY_STOCK`, …); the agnostic tokens stay legal (and required) on columns and functions. `register_currency` derives the `…_PER_SQUARE_METER_FLOW` variants so every currency-dimensioned kind is declarable.

### Literals

Strict literal handling is tested: a bare literal added to a unit-carrying value fails (must be tagged); a purely multiplicative literal needs no tag.

## Test plan

- [x] All CI green (tests on py311–py314 incl. jax across Linux/macOS/Windows; ty; codecov)


src/ttsim/interface_dag_elements/unit_checks.py
```diff
@@ -0,0 +1,998 @@
...
+        if isinstance(obj, ColumnObject | ParamFunction)
+        and node_is_boolean(qname=qname, obj=obj)
+    }
+    errors: list[str] = []
+    for qname, obj in env.items():
+        # Only these two have a scalar body written by a human; everything
+        # else is generated and unit-assigned by construction.
+        if not isinstance(obj, PolicyFunction | ParamFunction):
+            continue
+        if qname in boolean_nodes or qname not in resolved_units:
```
hmgaudecker: **Boolean-returning bodies are skipped, so the comparisons inside them go unchecked.** A body comparing `CURRENCY/month` with `CURRENCY` passes the environment check here (calling `_verify_one_body` directly catches it). Most mettsim eligibility functions are in this blind spot, and the miss can change decisions, not just levels. Running boolean bodies through the explorer and accepting only a dimensionless final result would close it.

src/ttsim/interface_dag_elements/unit_checks.py
```diff
@@ -0,0 +1,998 @@
...
+            continue
+        if not obj.verify_units:
```
hmgaudecker, line 521: **`verify_units=False` leaves the incoming edges unchecked.** There is no independent input-side contract, so an opted-out body's incoming edges go unchecked — contrary to GEP-10's 'edges are still checked'. Relatedly `ParamMappingObject` exposes only `output_unit` (≈L184), so a schedule's domain (`input_unit`) is never bound to the x argument: a `HECTARES` schedule can be evaluated at a currency input. Per-argument contracts, or a narrower claim, would match behavior.

src/ttsim/interface_dag_elements/unit_checks.py
```diff
@@ -0,0 +1,998 @@
...
+        return self._wrap(abs(self.q))
+
+
+def _wrap_for_dry_run(value: Any, explorer: _PathExplorer) -> Any:  # noqa: ANN401
+    """Wrap unit-carrying representative values; pass framework args through.
+
+    Quantities (and the leaves of dict-param trees) become ``_DryRunQuantity`` so the
+    explorer controls branches on them; ``xnp``/``num_segments``/… stay raw.
+    """
+    if isinstance(value, pint.Quantity):
```
hmgaudecker, line 1322: **An opaque return slips through without an opt-out.** `_verify_one_body` only acts on a non-dimensionless `pint.Quantity`; a dataclass/tuple return is neither checked nor rejected — a converter returning e.g. `BountyPerOrc` passes without `verify_units=False`, and a dataclass declared `YEARS` whose field carries `CURRENCY` is accepted. Requiring `verify_units=False` (or recursing into supported containers) for any non-scalar/non-Quantity return would close it.

src/ttsim/interface_dag_elements/unit_checks.py
```diff
@@ -0,0 +1,998 @@
...
+    result (e.g. an early ``return 0.0`` guard) falls back to the declaration on
+    that path. A run that *raises* — a body using a lookup table, a piecewise
+    polynomial, ``join``, or a raw ``xnp`` op the dry-run cannot evaluate — is
+    reported as needing an explicit ``verify_units=False`` opt-out (callers reach
+    this only for bodies that have not already opted out).
+    """
+    explorer = _PathExplorer()
+    paths = 0
+    while True:
+        if paths >= _MAX_PATHS:
```
hmgaudecker, line 1429: **Truncated path coverage returns success instead of failing.** At the 1024-path cap this only warns and returns success, so a wrong-unit branch first reached at path 1025 passes silently (the 64-decisions/run cap correctly forces an opt-out; this one doesn't). The explorer also records booleans but no path constraints, so it can explore infeasible `x>y ∧ x<=y` combinations and false-positive. Hard-failing/opting out on truncation would fix the silent miss.

src/ttsim/interface_dag_elements/unit_checks.py
```diff
@@ -0,0 +1,998 @@
...
+        self._explorer = explorer
+
+    def _wrap(self, q: Any) -> _DryRunQuantity:  # noqa: ANN401
+        return _DryRunQuantity(q=q, explorer=self._explorer)
+
+    def _controlled_bool(self) -> _DryRunQuantity:
+        # A dimensionless, explorer-controlled stand-in for a comparison result.
+        return self._wrap(UNIT_REGISTRY.Quantity(1.0, ""))
+
+    def _fail_if_other_unit_is_not_equivalent(self, other: Any, op: str) -> None:  # noqa: ANN401
```
hmgaudecker, line 1106: **Bare non-zero literals in comparisons silently carry a dimension.** `_fail_if_other_unit_is_not_equivalent` is lenient for all bare literals, not just `0`, so `wealth > 1_000_000` / `age < 33` are accepted though the literals implicitly carry `CURRENCY`/`YEARS`. Permitting only literal `0`, and rejecting non-zero bare literals in ordering/min/max unless tagged or promoted to a parameter, would catch these.

hmgaudecker (commented):

src/ttsim/interface_dag_elements/policy_environment.py
```diff
@@ -184,14 +303,36 @@ def _get_one_param(
...
+        cleaned_spec["value"] = _scale_numeric_leaves(
+            value=cleaned_spec["value"],
+            factor=_currency_conversion_factor_for_token(
+                raw_token=cleaned_spec.get("unit"), run_currency=currency
+            ),
+        )
         return ScalarParam(**cleaned_spec)
-    if param_type == "dict":
-        return DictParam(**cleaned_spec)
+    if param_type in ("dict", "require_converter"):
```
hmgaudecker, line 313: **`require_converter` values are converted by a single uniform factor, which is wrong for polynomial coefficients.** `require_converter` shares this flat-dict path: `_dict_param_value_in_run_currency` → `_scale_numeric_leaves` multiplies every numeric leaf by the same currency factor of the single `unit:` token. Correct for flat amounts (rent-cap lookups), but wrong for a structured converter with polynomial coefficients — contrast `_piecewise_param_value_in_run_currency` (L266), which correctly applies `f_out/f_in**j`. gettsim stores the Einkommensteuertarif as `require_converter`; with `f_in=f_out=0.25`, `{0,1,0.5}`→`{0,0.25,0.125}` yields 84.375 vs the correct 1275 at x=25. EUR-only today (factor 1). Rejecting non-trivial currency conversion of a require_converter unless its converter declares uniform leaves would close it; the single `unit:` token also can't type a mapping's domain≠codomain.

hmgaudecker (commented):

src/ttsim/interface_dag_elements/unit_checks.py
MImmesberger: Fixed — boolean-returning bodies are no longer skipped. Dropped the `boolean_nodes` clause from the env check, so a boolean body is now dry-run like any other (declares DIMENSIONLESS, full symmetry) and the comparisons inside eligibility functions are checked. Added `&`/`|`/`^`/`~` to `_DryRunQuantity` (screening both operands for dimensionless) so they run rather than forcing opt-outs; mettsim's one un-dry-runnable boolean (`in_same_household_as_recipient`, uses `join`) opts out explicitly. Pushed.

src/ttsim/interface_dag_elements/unit_checks.py
MImmesberger: Fixed the silent pass: hitting the path cap now returns the opt-out-required error (symmetric with the per-run decisions cap) instead of warn-and-succeed. On the infeasible-path point — decided to keep syntactic branch-combination coverage (no path-constraint tracking) and make the wording honest: the GEP now says 'syntactically reachable branch combinations' and notes the cap opts out. Pushed.

src/ttsim/interface_dag_elements/unit_checks.py
MImmesberger: Fixed — `_verify_one_body` now demands `verify_units=False` for any return that is neither a `pint.Quantity` nor a plain numeric/bool scalar (no recursion). `bounty_per_orc` opts out accordingly; the check also correctly caught a few `not_required` test fixtures returning columns/lists, now opted out. GEP opt-out paragraph extended. Pushed.

src/ttsim/interface_dag_elements/unit_checks.py
MImmesberger: Fixed — the ordering comparisons reject a bare non-zero literal when the quantity is non-dimensionless (`0` stays the lone inline literal; dimensionless self stays lenient; `min`/`max` route through the same path). Scope is ordering only — `+`/`-` keep the lenient `0`-guard handling. mettsim's `age < 33` became an `age_of_majority` (YEARS) parameter. GEP catch-bullet amended. Pushed.

src/ttsim/interface_dag_elements/policy_environment.py
MImmesberger: Fixed. A `require_converter` no longer always routes through the uniform-scaling path. It now declares one of two honest shapes:

- a single homogeneous `unit:` (every numeric leaf the same currency) → scaled uniformly, as before; or
- `input_unit:` / `output_unit:` axes when its converter produces a function-like value (a schedule / lookup) → the raw blob is left untouched and the converter's **typed output** is converted per axis, so an order-`j` coefficient scales by `f_out / f_in**j` (slope invariant, quadratic by `1/f_in`), reusing the same per-axis machinery as the `piecewise_*` params.

A structure that *mixes* a currency with non-currency numbers (the Einkommensteuertarif's slopes/intercepts, or a `satz` bundled with an age bracket) has no surface to declare a unit per leaf in a single blob, so it is split into separate homogeneous parameters. As a fail-loud backstop, a homogeneous (single-`unit:`) converter found to produce a function-like value is now rejected at build time, with the error pointing at the per-axis declaration.

Implemented in pint-120 (`b57e9ccb`); worked example (the king's levy — a quadratic schedule) + the polynomial-per-order regression test in pint-121 (`80be1f67`).

src/ttsim/interface_dag_elements/unit_checks.py
MImmesberger: Resolved as a documented known limitation — decided against building the domain-binding stub.

The "incoming edges unchecked" part isn't a real gap in practice: every *other* body in the DAG is dry-run, so an opted-out body's inputs are the separately-verified declared outputs of their producers. The only way a wrong unit could reach an opt-out body unseen is if two opt-out bodies fed each other directly (e.g. back-to-back `piecewise_polynomial`s), which doesn't occur here. The genuine residual is that the opted-out body itself is never dry-run, so a schedule's `input_unit` domain is never bound to the argument it's evaluated at — but `piecewise_polynomial` is a single standardized primitive, so the risk is small.

Fixed the overstated comment here (folded into the edge-check commit, now `9fa15960`) to say exactly that, and narrowed the matching GEP claim + added a "Known limitation" note (gettsim `0422bb5a`).

MImmesberger:

PR #125, opened 2026-06-03
pint units: currency knob + Layer-2 boundary assertions (#120)

### LINKED ISSUE #120: pint units: currency knob + Layer-2 boundary conversion
## What to build

Add the currency main() input (default euro) as an interface-DAG node setting the currency of input data and output. At build time convert each param from its declared currency to the run currency and bake the factor; functions stay currency-agnostic. Add the Layer-2 boundary handling: users MAY attach pint Quantity arrays to input columns; at the canonicalization boundary the tag's currency is **converted** to the run currency (a DM/silver_penny-tagged column feeds a euro/castar run, rescaled at the boundary), then stripped to a bare array. A **strict period guard** requires the tag's flow period to match the column's GEP-1 time suffix exactly (a `_m` column needs a `/month` tag) — read off the qname, so no declared unit is needed — so a contradictory period fails loudly. An unknown-token tag fails loudly too. Demonstrate in mettsim with two registered currencies.

**Note (per GEP-10 review, 2026-06-15):** the boundary **converts** currency rather than asserts, reversing the original "tags never convert" plan, and adds the strict period guard above. Validating the tag against the column's *declared* unit (to reject a wrong *dimension*, e.g. a currency on an age column) is deferred — it would need the declared unit threaded to the boundary (the earlier "blocked by a cycle" framing was wrong: only the *spec* env depends on the data, not `policy_environment`).

## Acceptance criteria

- [ ] currency main() node, default euro; output in the run currency
- [ ] params converted source-to-run currency at build; functions untouched
- [ ] pint-tagged inputs have their currency converted to the run currency and stripped at the boundary; a period that disagrees with the column's time suffix, or an unknown-token tag, fails loudly
- [ ] mettsim demonstrates a second-currency run
- [ ] tests green

## Blocked by

- https://github.com/ttsim-dev/ttsim/issues/118

MImmesberger:
Closes #120. Fourth PR of the GEP 10 series. **Stacked on #124 (#119)** — review/merge that first.

### Currency knob
A new `currency` interface node sets the currency of the input data and of the output. It **defaults to the registered base currency** and is overridable via `main(currency=...)`. Functions stay currency-agnostic. (It is anchored on `backend` in the interface DAG so an argument-less node isn't treated as a missing root input.)

### Token-driven build-time conversion
A parameter's unit declaration names the currency its numbers are written in (`SILVER_PENNY_STOCK`, `CASTAR_FLOW`, …). At environment-build time every currency-denominated number is converted from that denomination to the run currency and the factor baked in — historical values are stored in their legal currency rather than hand-converted. There is **no separate `source_currency:` key**: one declaration mechanism.

Conversion covers **every parameter shape**:
- scalar values (× the currency factor);
- piecewise schedules per axis — interval bounds by the input factor, intercepts by the output factor, order-*j* coefficients by `f_out / f_in**j` (income-schedule slopes are invariant; area-schedule slopes carry the full output factor);
- lookup-table values by the output factor (a concrete-currency *input* axis is rejected — integer-keyed);
- dict params leaf by leaf.

A **dated entry may restate the unit field(s)** — this is how a currency changeover within one parameter's history is written (old entries in the legacy currency, entries from the reform date in the new one). `updates_previous` cannot cross a changeover.

`currency` is threaded through `policy_environment` → `_active_param_objects` → `_get_one_param`; conversion happens on the cleaned spec before the param is constructed.

### Layer-2 boundary
A user MAY attach a pint `Quantity` to an input column. At the GEP-9 canonicalisation boundary the tag's **currency is converted** to the run currency (a `silver_penny`/DM-tagged column feeds a `castar`/euro run, rescaled at the boundary) and then stripped to a bare array; its area is left untouched. The tag's **period must match the column's GEP-1 time suffix exactly** (a `_m` column needs a `/month` tag, an unsuffixed column a tag with no period) — read off the qname, so no declared unit is needed — so a contradictory period (a `_m` column tagged `/year`) fails loudly instead of being silently mis-scaled; the flow period is the *denominator* time unit, so an age (`year` numerator) is not flagged. A bare column is taken to already be in the run currency. A tag built from a unit token TTSIM does not know also fails loudly. Validating the tag's *dimension* against the column's declared unit (reject a currency on an age column) is deferred to future work.

> **Update (per GEP-10 review):** the boundary now **converts** currency rather than asserts (`strip_input_quantity_at_boundary`), reversing the original "tags never convert", and adds a **strict period guard** (the tag's flow period must equal the column's `_y`/`_m` suffix exactly). Drive-by fix: `_canonicalize_input_dtype`'s annotation now admits `pint.Quantity`, which beartype had been rejecting — the boundary was wired but never exercised end-to-end with a real tag. Covered by boundary unit tests + an end-to-end test through the `processed_data` interface node (numpy + jax).

### mettsim two-currency demo
mettsim registers **castar** (base) and **silver_penny = castar / 4**. A `currency="silver_penny"` run converts a castar-denominated threshold to four times its magnitude at build time, while a default (castar) run keeps it.

## Acceptance criteria

- [x] `currency` main() node, default base currency; output in the run currency
- [x] params converted to the run currency at build (scalar/piecewise/lookup/dict); functions untouched
- [x] per-entry changeover overrides; `updates_previous` cannot cross one
- [x] pint-tagged inputs have their currency converted to the run currency and stripped at the boundary; a period that disagrees with the column's time suffix, or an unknown-token tag, fails loudly
- [x] mettsim demonstrates a second-currency run

## Verification

All CI green (tests on py311–py314 incl. jax across Linux/macOS/Windows; ty; codecov). `tests/test_currency_knob.py` covers scalar / piecewise / lookup / changeover conversion.


hmgaudecker (commented):

src/ttsim/interface_dag_elements/unit_checks.py
```diff
@@ -276,6 +276,74 @@ def fail_if_environment_units_are_inconsistent(
...
+            with its declared unit. All offending columns are reported together.
+    """
+    errors: list[str] = []
+    for qname, tag in input_units.items():
+        expected = resolved_units.get(qname)
+        if not isinstance(expected, pint.Unit):
+            # Column without a (scalar) declared unit — e.g. absent from the
+            # environment, or a dict parameter. Nothing to check here.
+            continue
+        tag_dim = UNIT_REGISTRY.Quantity(1.0, tag).dimensionality
```
hmgaudecker: **A tagged input of a compatible but differently-scaled unit is accepted and then mis-scaled.** This fail node compares only dimensionality, while the live strip path (`processed_data.py:53` → `strip_input_quantity_at_boundary`) converts currency only and leaves area/time scale untouched. So a `HECTARES` input tagged `10_000 m²` passes and enters the runtime as the bare number 10000 instead of 1; `12 month` for a `YEARS` input becomes 12 instead of 1 — a silent level error that can move brackets/eligibility. Converting to the expected unit before stripping, or rejecting compatible-but-non-equivalent non-currency tags, would close it.

MImmesberger: Fixed in `1001a608` (pint-120). `fail_if_input_units_are_inconsistent` now compares a tag to its declared unit for *equivalence*, not merely shared dimensionality. It factors out the two axes the boundary handles separately — the currency (converted to the run currency at the boundary) and the flow period (screened against the name suffix by the dedicated period guard) — via a new `unit_residual_excluding_currency_and_flow_period` helper (pint-117), then requires `units_are_equivalent` on the residual. So a `HECTARES` column tagged `m²` (a 10,000-fold level error) or a `YEARS` age tagged `month` (12-fold) is now rejected at build time instead of being silently mis-stripped. Tests added in `test_input_units.py` (including one asserting a `_m`/per-year period mismatch is left to the suffix guard, not double-reported here).

MImmesberger: Follow-up: the GEP prose now matches the shipped behaviour too (gettsim `2dc80ba7`). The Layer-2 description had still called this "the dimension check"; it now states the boundary requires each tagged input to be *equivalent* to its resolved unit once currency (converted by the strip path) and a flow's reference period (owned by the suffix guard) are divided out — so a `HECTARES` column tagged `m²` or a `YEARS` input tagged in months is rejected rather than silently mis-scaled.

MImmesberger:

PR #129, opened 2026-06-06
Remove dead SIM108 ignore for middle_earth; showcase ternaries in mettsim

MImmesberger:
Three related changes, following the discovery in https://github.com/ttsim-dev/gettsim/pull/1178#discussion_r3366916727 that the vectorizer handles ternary expressions (`ast.IfExp` → `xnp.where` via `_ifexp_to_call`):

1. **Remove the dead SIM108 ignore** for `src_mettsim/mettsim/middle_earth/**/*.py`. The ignore never suppressed anything: SIM108 only flags *assignment-style* if/else (`if c: x = a / else: x = b`; verified on ruff 0.13.0, stable and preview alike), while mettsim's if/else blocks are *return-style* — which no ruff rule converts to a ternary. The single assignment-style block (orc bounty) exceeds the line limit, where SIM108 auto-skips.

2. **Convert two return-style if/else to ternaries** (wealth tax, child tax credit) so that mettsim showcases ternaries in real policy functions — these are swept by `test_convertible` for all policy years. Only sites where the ternary is a genuine short one-liner are converted: mettsim doubles as test material for the vectorizer, so it deliberately keeps exercising both supported styles. Vectorization of ternaries was already unit-tested via `f2`/`f6`/`f7`/`f10` in `tests/tt/test_vectorization.py`.

3. **Narrow mettsim's blanket `RET` ignore to `RET505`.** Only RET505 (superfluous else after return) conflicts with the vectorizer: its autofix produces the early-return form (`if c: return a` + bare `return b`) that `_if_to_call` rejects by design (see the `g3` fixture). The rest of the RET family passes clean on mettsim and its autofixes are vectorizer-safe, so it is now enforced.

The SIM108 ignore for `tests/tt/test_vectorization.py` stays: its statement-style if/else functions are intentional fixtures for the GAST converter.

Companion sweep for gettsim's policy code: ttsim-dev/gettsim#1195.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -214,7 +214,6 @@ per-file-ignores."src_mettsim/mettsim/middle_earth/**/*.py" = [
   "PLR1716", # Boolean chained comparison
   "PLR2004", # Magic value comparison
   "RET",     # Unnecessary return logic
-  "SIM108",  # Use ternary operator
 ]
 per-file-ignores."src_mettsim/tests_middle_earth/*.py" = [
   "ANN",     # Missing type annotations
@@ -240,7 +239,7 @@ per-file-ignores."tests/tt/test_vectorization.py" = [
   "PLR1714", # Repeated equality comparison
   "PLR1716", # Boolean chained comparison
   "RET",     # Unnecessary return logic
-  "SIM108",  # Use ternary operator
+  "SIM108",  # Statement-style if/else blocks are intentional fixtures for the vectorizer
 ]
 pydocstyle.convention = "google"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -209,9 +209,6 @@ per-file-ignores."src/ttsim/typing.py" = [
   "PYI051", # Redundant literal union
 ]
 per-file-ignores."src_mettsim/mettsim/middle_earth/**/*.py" = [
-  "E721",    # Type comparison using ==
-  "PLR1714", # Repeated equality comparison
-  "PLR1716", # Boolean chained comparison
   "PLR2004", # Magic value comparison
   "RET505",  # Vectorizer requires both if/else branches; it rejects the early-return form
 ]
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -240,7 +239,7 @@ per-file-ignores."tests/tt/test_vectorization.py" = [
   "PLR1714", # Repeated equality comparison
   "PLR1716", # Boolean chained comparison
   "RET",     # Unnecessary return logic
-  "SIM108",  # Use ternary operator
+  "SIM108",  # Statement-style if/else blocks are intentional fixtures for the vectorizer
```
hmgaudecker: Fine, but we should add a test for the new behaviour, or is it already in place?

I'm amazed we don't have a single instance of an if/else block in METTSIM.

hmgaudecker (approved):
Thanks!

pyproject.toml
MImmesberger: We have if/else blocks, the SIM108 rule is about how they're written up (and we already wrote everything in a compliant way).

I added two oneline if/else conditions to the code and tightened the RET hook (we can allow all codes but RET505 I believe).

AFAICT there is no rule that disallowed these oneliners. Sorry, should have made that clear earlier.

pyproject.toml
```diff
@@ -240,7 +239,7 @@ per-file-ignores."tests/tt/test_vectorization.py" = [
   "PLR1714", # Repeated equality comparison
   "PLR1716", # Boolean chained comparison
   "RET",     # Unnecessary return logic
```
hmgaudecker:
```suggestion
  "RET505",  # Vectorizer requires both if/else branches; it rejects the early-return form
```
?

hmgaudecker (approved):
Excellent, thanks!

pyproject.toml
hmgaudecker: I wonder whether we need all that duplication in the first place, though.

hmgaudecker:
:innocent: deleted a message where my own bad memory fooled me, sorry for the email noise, @MImmesberger

PR #131, opened 2026-06-06
Split raw_results.columns into columns_with_internal_p_ids / columns_with_original_p_ids

MImmesberger:
Closes #130.

- Split `raw_results.columns` into `raw_results.columns_with_internal_p_ids` (computable from `processed_data` alone, endogenous `p_id_*` columns hold internal indices) and `raw_results.columns_with_original_p_ids` (the reverse-translation from #108, consumed by `results.tree`, so all `results.*` outputs keep returning user-space `p_id`s).
- Both nodes are exposed via `MainTarget` and the `RawResults` main arg.
- New `src_mettsim/tests_middle_earth/test_jittability.py` exercises the `processed_data`-only path over all mettsim column functions.
- Breaking: the target `raw_results.columns` no longer exists; downstream switches to `columns_with_internal_p_ids` (ttsim-dev/gettsim#1196).

### DIFF
--- a/tests/interface_dag_elements/test_raw_results.py
+++ b/tests/interface_dag_elements/test_raw_results.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 from pathlib import Path
 
 import ttsim.interface_dag_elements
-from ttsim.interface_dag_elements.interface_node_objects import InterfaceFunction
 from ttsim.interface_dag_elements.orig_policy_objects import load_module
 
 # The package claw rewrites normally-imported `interface_dag_elements` modules,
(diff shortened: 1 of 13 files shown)


tests/interface_dag_elements/test_raw_results.py
```diff
@@ -301,11 +293,17 @@ def test_params_returns_various_value_types():
...
-    assert columns.dependencies == {
+def test_columns_with_internal_p_ids_dependencies():
+    assert columns_with_internal_p_ids.dependencies == {
         "labels__root_nodes",
         "processed_data",
         "tt_function",
+    }
+
+
+def test_columns_with_original_p_ids_dependencies():
```
hmgaudecker, line 304: Should we add the corresponding `test_columns_with_internal_p_ids_dependencies`, swapping `input_data__flat` and `input_data__sort_indices` for `processed_data` ? 

I guess that would be the direct test for the failure mode that surfaced, no?

hmgaudecker (approved):
Sweet!

tests/interface_dag_elements/test_raw_results.py
MImmesberger: Already exists right above this one!

PR #133, opened 2026-06-11
Replace hand-written ty machinery with the official ty-pre-commit hook

hmgaudecker:
## Summary

ty now runs only as pre-commit hooks, under both backends, replacing the dedicated pixi `type-checking`/`type-checking-jax` features, environments, and tasks (`pixi run ty` / `pixi run ty-jax` no longer exist).

- **numpy run**: the official [astral-sh/ty-pre-commit](https://github.com/astral-sh/ty-pre-commit) hook with `args: [--no-project]`. uv creates no `.venv`/`uv.lock` in this pixi-managed repo and resolves no dependencies itself (its default mode fails on the pygraphviz source build anyway); ty resolves third-party imports from the pixi `py314` environment via `tool.ty.environment.python` in `pyproject.toml`.
- **jax run**: a minimal local hook running the same pinned ty with `--python .pixi/envs/py314-jax`. It inherits the entire `tool.ty` configuration — same rules, only the environment differs. Verified that jax genuinely resolves in this run (passes even with jax dropped from `analysis.allowed-unresolved-imports`).

The jax variant cannot use the official hook today: `uv check` forwards nothing to ty (no `--python`, no `-c`, no trailing args), and ty rejects conda-style envs handed over via `VIRTUAL_ENV` (no `pyvenv.cfg`). The local hook's `ty==0.0.49` pin carries a comment to keep it in sync with the hook rev, since `prek auto-update` bumps only the rev.

## Other changes

- The `run-ty` CI job installs `py314`/`py314-jax` and runs the hooks via prek, so the pre-commit config is the single definition of how ty runs everywhere.
- Both ty hooks are skipped on pre-commit.ci (no network at hook runtime, no pixi environments).
- The existing test environments provide everything ty needs (including type stubs), so the dedicated type-checking environments are gone — including their ~1800 lines in `pixi.lock`. pixi 0.70.2 does not prune removed environments from the lock; the stale blocks and orphaned `ty` package entries were removed manually and the result validated with `pixi lock`.
- AGENTS.md: verification is now two steps; `prek run --all-files` covers type checking under both backends.
- `.ai-instructions` is re-pinned to c2701d4 (the merged boilerplate: real tiering, distilled modules, ty-pre-commit guidance) — matching the rest of the gettsim family.
- **ty bumped 0.0.48 → 0.0.49** (hook rev + local jax pin), matching the `.ai-instructions` boilerplate and the rest of the family. 0.0.49 flagged a non-exhaustive `shape_id` if/elif chain in `tests/test_end_to_end.py` as a possibly-unresolved reference; fixed with an `else` guard that raises on an unknown `shape_id`.

## Test plan

- [x] `pixi run -e py314-jax prek run --all-files` — all hooks pass, including both ty variants (numpy + jax) on ty 0.0.49
- [x] `pixi run --locked -e py314 prek run ty --all-files` and `... ty-jax --all-files` (exact CI commands)
- [x] `pixi run --locked -e py314-jax tests -n 7` — 1119 passed, 42 skipped
- [x] `pixi lock` — lock file validates after pruning


MImmesberger (approved):

AGENTS.md
```diff
@@ -26,18 +26,19 @@ pixi run -e py314 pytest tests/test_main.py
 # Run a single test
 pixi run -e py314 pytest tests/test_main.py::test_function_name -v
 
-# Type checking with ty
-pixi run -e ty ty
+# Type checking with ty (runs as pre-commit hooks against the py314 and
+# py314-jax environments; requires both to be installed)
+pixi run -e py314 prek run ty --all-files
```
MImmesberger:
```suggestion
prek run ty --all-files
```
Same below.

hmgaudecker: Done in 561eea95 — dropped the `pixi run -e py314` prefix on both the `ty` and `ty-jax` lines.

PR #139, opened 2026-06-27
pint units (GEP 10) · infra 2/3: currency & boundary

MImmesberger:
Concrete-currency handling and the Layer-2 boundary: the currency knob, unit-annotated input/output boundary checks, per-axis `require_converter` currency conversion, and scalar-param/`reference_level` constraints.

> Part of the GEP-10 stack. Per the worked-example convention, only the top PR (`gep10-annotate-mettsim`) is green; these infra PRs are red by design — units become *mandatory* here but the worked example that satisfies them is annotated at the top.


MImmesberger:
Collapsed into a single infra PR #138. The three-way infra split introduced the token-based unit model in #138 and replaced it with compositional units here/in #140, so the same lines were re-touched across PRs (units.py ~1379 lines, unit_checks.py ~671, …). #138 now contains the whole GEP-10 infra as one final-form diff (no token model anywhere); #141 (worked example) is retargeted onto it.

PR #140, opened 2026-06-27
pint units (GEP 10) · infra 3/3: compositional units + T8

MImmesberger:
The fully-spelled compositional unit vocabulary (fluent builder, `[hours]` as its own dimension, leveled booleans) and the T8 aggregation-level invariant (a column's `_xx` suffix and its unit's grouping level always in sync; group aggregations resolve to their target level).

**Head-count base — `PERSON_COUNT`.** The head-count base is spelled `Unit.PERSON_COUNT` / `PERSON_COUNT_PER_<level>`, distinct from the implied individual `person` grouping *level* (`PER_PERSON`). The token resolves to the `[person]` dimension, which a per-person `/[person]` denominator cancels — so `mean_per_person × head_count` types as a total.

**Leveled booleans + implied person leaf** are finalized on the compositional column/param path here: a boolean is a leveled `DIMENSIONLESS` (bare for the person leaf, `DIMENSIONLESS_PER_<group>` for a group), routed through `resolve_compositional_column_unit` via an `is_boolean` seam; extensive *parameters* imply the person leaf just like columns; `ANY`/`ALL` mint `DIMENSIONLESS_PER_<target>`; spelling the person leaf (`PER_PERSON`) on the builder is rejected (it is implied, never spelled).

> Part of the GEP-10 stack. Per the worked-example convention, only the top PR (`gep10-annotate-mettsim`) is green; these infra PRs are red by design — units become *mandatory* here but the worked example that satisfies them is annotated at the top.


MImmesberger:
Collapsed into a single infra PR #138. The three-way infra split introduced the token-based unit model in #138 and replaced it with compositional units here/in #140, so the same lines were re-touched across PRs (units.py ~1379 lines, unit_checks.py ~671, …). #138 now contains the whole GEP-10 infra as one final-form diff (no token model anywhere); #141 (worked example) is retargeted onto it.

PR #151, opened 2026-08-17
GEP 10: add auditable unit validation to TTSIM

hmgaudecker:
## Summary

This PR implements the TTSIM unit checker specified by GEP 10. It is stacked on [#150](https://github.com/ttsim-dev/ttsim/pull/150).

The implementation:

- adds public `TTSIMUnit.COUNT` and `TTSIMUnit.INDICATOR` declarations for hand-written counts and yes/no indicators; both remain dimensionless in Pint arithmetic while preserving the narrow semantic evidence needed for documented group calculations;
- checks supported policy expressions involving arithmetic, comparisons, conditions, `xnp.where`, joins, aggregations, reductions, schedules, mappings, structured parameter fields, calendar values, and local unit casts;
- restricts group arithmetic to the documented cases for totals, head counts, indicators, group statistics, and person-level allocations;
- validates declarations and function bodies across every relevant policy-date regime;
- exposes `MainTarget.unit_checks.validation_report`, with separate classifications for resolved declarations, checked bodies, checked aggregations, generated rules, local casts, whole-body opt-outs, unsupported bodies, and other unchecked bodies; and
- expands the TTSIM and METTSIM regression tests for both NumPy and JAX execution.

Ordinary environment validation fails closed when a unit-relevant operation cannot be checked. The diagnostic report makes every exception visible without treating a declaration check as proof of arbitrary Python, data alignment, legal correctness, or the economic meaning of otherwise dimensionless values.

## NumPy compatibility

A separate commit makes backend-change detection tolerate removed NumPy scalar attributes that still appear in `dir()`, such as `itemset`. Readable array-valued attributes are still detected; inaccessible tombstone attributes are ignored.

## Related PRs

- [GETTSIM #1220](https://github.com/ttsim-dev/gettsim/pull/1220) specifies the behavior in GEP 10.
- [GETTSIM #1212](https://github.com/ttsim-dev/gettsim/pull/1212) applies the declarations and validation to the German policy system.

## Verification

- `pixi run -e py314-jax prek run --all-files`
- `pixi run -e py314-jax tests -n 7`
- GitHub Actions covers NumPy on Python 3.11–3.14 and NumPy/JAX on Linux, macOS, and Windows.


src/ttsim/_quantity_kinds.py
```diff
@@ -0,0 +1,185 @@
...
+    resolve_kind_of_column_function,
+)
+from ttsim.tt.units import QuantityKind
+
+_INTEGER_KINDS = frozenset({ResolvedKind.INT_SCALAR, ResolvedKind.INT_COLUMN})
+
+# A direct integer only qualifies when its documentation actually says it is a
+# count. English and German terms cover the policy packages maintained here; the
+# integer annotation by itself is deliberately insufficient.
+_COUNT_MEANING = re.compile(
```
MImmesberger: This seems a bit dangerous to me. We can merge and see whether any issues come up (the
new features are quite orthogonal to the rest so this wouldn't do much harm).

Alternatively, I would propose to add user-facing units like count or identifier. Users
would type, for example, `TTSIMUnit.COUNT`. Internally, we immediately rewrite this into
`unit=TTSIMUnit.DIMENSIONLESS` and `kind=QuantityKind.COUNT`.

The advantage would be that we retain the meaning of a dimensionless unit internally
while having an interface that is easier to understand (I hope). What do you think?

hmgaudecker: Implemented this approach in 33658c42. The public API now has TTSIMUnit.COUNT and TTSIMUnit.INDICATOR (including COUNT_PER_* / INDICATOR_PER_* in YAML); each normalizes immediately to the dimensionless Pint base plus internal checker metadata. I also removed the public QuantityKind side channel and all name/qname/docstring heuristics. The explicit declaration propagates through scalars, mapping leaves, schedule axes and outputs, structured fields, and casts, while generated COUNT, Boolean SUM, ANY, and ALL retain automatic evidence. The full NumPy and JAX suites pass, and the TTSIM PR CI is green.

